PR #33, opened 2016-12-14
fix pre-commit hook so displays unformatted files

jdhenke:
before, if unformatted files existed, pre-commit hook would silently exit non-zero with no explanation.

this commit fixes so that if `./godelw format` returns non-zero, so does the pre-commit, and if there are unformatted files, it logs those as well.

here's the new code with syntax highlighting:

```bash
gofiles=$(git diff --cached --name-only --diff-filter=ACM | grep '\.go$')
[ -z "$gofiles" ] && exit 0

unformatted=$(./godelw format -l runAll $gofiles)
exitCode=$?
[ "$exitCode" -eq "0" ] && exit 0

if [ -n "$unformatted" ]; then
  echo "Unformatted files exist -- run ./godelw format to format these files:"
  for file in $unformatted; do
    echo "  $file"
  done
fi

exit $exitCode
```

### DIFF
--- a/cmd/githooks/hooks.go
+++ b/cmd/githooks/hooks.go
@@ -30,16 +30,16 @@ var hooks = map[string]string{
 
 unformatted=$(./godelw format -l runAll $gofiles)
 exitCode=$?
-[[ $exitCode -ne 0 ]] && exit $exitCode
-[ -z "$unformatted" ] && exit 0
+[ "$exitCode" -eq "0" ] && exit 0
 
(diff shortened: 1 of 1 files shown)


nmiyake:
Thanks for the PR -- this looks good! I'm going to add a commit on top of this that adds an integration test for this and will merge that.

nmiyake:
Closing in favor of #35 (builds on top of this commit to add integration tests and `#!` to script)

PR #45, opened 2017-01-26
Initial impl of clean

andybradshaw:
Took a stab at making a `clean` command. Tested locally against the some repositories that I have, but will add some unit tests.

### DIFF
--- a/apps/distgo/cmd/clean/clean.go
+++ b/apps/distgo/cmd/clean/clean.go
@@ -1,0 +1,76 @@
+// Copyright 2016 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clean
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
+
+	"github.com/palantir/godel/apps/distgo/params"
+)
+
+func Clean(products []string, cfg params.Project) error {
+	// Only remove default output directories if no products specified
+	if len(products) == 0 {
+		if err := remove(cfg.BuildOutputDir); err != nil {
+			return errors.WithStack(err)
+		}
+		if err := remove(cfg.DistOutputDir); err != nil {
+			return errors.WithStack(err)
+		}
+	}
+
+	err := cleanProducts(products, cfg)
+	if err != nil {
+		return errors.Wrap(err, "failed to clean products")
+	}
+
+	return nil
+}
+
+func cleanProducts(products []string, cfg params.Project) error {
+	// remove all products
+	if len(products) == 0 {
+		for name := range cfg.Products {
+			products = append(products, name)
+		}
+	}
+
+	for _, product := range products {
+		p := cfg.Products[product]
+		if err := remove(p.Build.OutputDir); err != nil {
+			return errors.WithStack(err)
+		}
+		for _, dist := range p.Dist {
+			if err := remove(dist.OutputDir); err != nil {
+				return errors.WithStack(err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func remove(dir string) error {
+	if dir != "" {
+		fmt.Printf("Removing %s\n", dir)
+		if err := os.RemoveAll(dir); err != nil {
+			return errors.Wrapf(err, "failed to remove dir %s", dir)
+		}
+	}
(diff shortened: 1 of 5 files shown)


apps/distgo/cmd/clean/clean.go
```diff
@@ -0,0 +1,88 @@
...
+		if err := remove(cfg.BuildOutputDir); err != nil {
+			return errors.WithStack(err)
+		}
+
+		if cfg.DistOutputDir == "" {
+			cfg.DistOutputDir = defaultDistOutputDir
+		}
+		if err := remove(cfg.DistOutputDir); err != nil {
+			return errors.WithStack(err)
+		}
```
nmiyake: Should there be a `return nil` at the end of this `if` block? As currently written, if `len(products) == 0`, this will remove both default directories and all project directories. Is that intended?

apps/distgo/cmd/clean/clean.go
```diff
@@ -0,0 +1,88 @@
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func remove(dir string) error {
+	if dir != "" {
+		fmt.Printf("Removing %s\n", dir)
```
nmiyake: Avoid using `fmt.Printf` because it's less modular/testable. Instead, pass a `io.Writer` through the functions and use `fmt.Fprintf` to print to the stream. At the top-level call in the CLI, you can then pass in the handle to the output writer provided by the context.

integration_test/integration_test.go
```diff
@@ -509,6 +509,140 @@ products:
...
+	err = os.MkdirAll(path.Join(testProjectDir, "bar"), 0755)
+	require.NoError(t, err)
+	err = ioutil.WriteFile(path.Join(testProjectDir, "bar", "bar.go"), []byte(src), 0644)
+	require.NoError(t, err)
+
+	gittest.CommitAllFiles(t, testProjectDir, "Commit files")
+	gittest.CreateGitTag(t, testProjectDir, "0.1.0")
+
+	execCommand(t, testProjectDir, "./godelw", "build")
+	fooBuild := execCommand(t, testProjectDir, "find", "foo/build", "-type", "f")
```
nmiyake: Would prefer for tests to use standard Go library commands from `io` and `ioutils` package for doing things like checking for existence. The functionality provided by things like `ls` and `find` can be written directly in Go, and would prefer to do that whenever possible (would like to minimize the use of `execCommand` to do only the things that have to be shelled out).

apps/distgo/cmd/clean/clean.go
```diff
@@ -0,0 +1,88 @@
...
+				return errors.WithStack(err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func remove(dir string) error {
+	if dir != "" {
```
nmiyake: Does this ever get called with `dir == ""`? Seems like that would be an error case/incorrect use of the function, in which case I think it would be better to return an error rather than silently no-oping.

apps/distgo/cmd/clean/clean.go
```diff
@@ -0,0 +1,88 @@
...
+	defaultDistOutputDir  = "dist"
+)
+
+func Clean(products []string, cfg params.Project) error {
+	// Only remove default output directories if no products specified
+	if len(products) == 0 {
+		if cfg.BuildOutputDir == "" {
+			cfg.BuildOutputDir = defaultBuildOutputDir
+		}
+		if err := remove(cfg.BuildOutputDir); err != nil {
```
nmiyake: In both the `remove` calls here and in `cleanProducts`, I believe that you'll run into an error/failure if the `build` and `dist` output directories are the same directory (since it will try to remove the same path twice).

Might be cleaner to make `remove` take a `[]string` instead of just `string` and then have the logic in this portion and the other portion build up a slice of unique paths to remove and then call the `remove` function with that (or have the `remove` function handle de-duping equivalent paths in the input).

integration_test/integration_test.go
andybradshaw: Sounds good to me. Wrote a little helper to replicate the find functionality, not sure if it's in the right place.

apps/distgo/cmd/clean/clean.go
```diff
@@ -0,0 +1,151 @@
...
+		if err != nil {
+			if strings.Contains(err.Error(), enotfound) {
+				return nil
+			}
+			return errors.WithStack(err)
+		}
+
+		fmt.Fprintf(stdout, "Removing %s\n", dir.path)
+		for _, file := range files {
+			if matchIdx := dir.pattern.FindStringIndex(file.Name()); matchIdx == nil {
```
nmiyake: In Go, it's more idiomatic to structure conditionals such that the main logic is as "unindented" as possible (makes reading the standard case more straightforward).

Fro example, in this case:

```go
for _, file := range files {
    if dir.pattern.FindStringIndex(file.Name()); matchIdx != nil {
        continue
    } else if !force {
        fmt.....
        continue
    }
    fmt.Fprtinf...
}
```

Obviously multiple different ways to do this, but this is the general principle.

nmiyake:
Would still like this functionality, but the PR is stale -- closing for now.

PR #52, opened 2017-02-04
Add optional init.sh reload

### LINKED ISSUE #51: distgo: opt-in to init.sh reload
`init.sh reload` is an optional command for sending a SIGHUP to the application. Unfortunately if you don't handle the SIGHUP, your process dies.

There should be `reload: true` option for distgo which includes the reload command. This will eliminate a reason that applications need to use templates.

bmoylan:
Fixes #51 

@nmiyake sorry the ticket was unclear, this is what I'm thinking

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/godel/52)

### DIFF
--- a/apps/distgo/cmd/dist/dist_test.go
+++ b/apps/distgo/cmd/dist/dist_test.go
@@ -926,6 +926,74 @@ daemon: true
 			},
 		},
 		{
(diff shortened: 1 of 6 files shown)


bmoylan:
ok ready to go

PR #93, opened 2017-04-02
Add project-version command

### LINKED ISSUE #92: distgo: `godel project-version`
Convenience command for printing the git version of the project that gets set to `version-var`. Shouldn't have any other output, the common usage will be
```sh
version=$(./godelw project-version)
```

nmiyake:
* Add "project-version" command to distgo that prints the version
  used for build/dist commands.
* Add "project-version" command to gödel that calls distgo command

Fixes #92

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/godel/93)

### DIFF
--- a/apps/distgo/app.go
+++ b/apps/distgo/app.go
@@ -23,6 +23,7 @@ import (
 	"github.com/palantir/godel/apps/distgo/cmd/build"
 	"github.com/palantir/godel/apps/distgo/cmd/dist"
 	"github.com/palantir/godel/apps/distgo/cmd/products"
+	"github.com/palantir/godel/apps/distgo/cmd/projectversion"
 	"github.com/palantir/godel/apps/distgo/cmd/publish"
 	"github.com/palantir/godel/apps/distgo/cmd/run"
 )
@@ -32,6 +33,7 @@ func App() *cli.App {
 	app.Name = "distgo"
 	app.Usage = "Build, run, test and publish products in a Go project"
 	app.Subcommands = []cli.Command{
+		projectversion.Command(),
 		products.Command(),
 		artifacts.Command(),
 		build.Command(),
--- a/apps/distgo/cmd/projectversion/cmd.go
+++ b/apps/distgo/cmd/projectversion/cmd.go
@@ -1,0 +1,34 @@
+// Copyright 2016 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package projectversion
+
+import (
+	"github.com/nmiyake/pkg/dirs"
+	"github.com/palantir/pkg/cli"
+)
+
+func Command() cli.Command {
+	return cli.Command{
+		Name:  "project-version",
+		Usage: "Prints the version of the project (based on Git information)",
+		Action: func(ctx cli.Context) error {
+			wd, err := dirs.GetwdEvalSymLinks()
+			if err != nil {
+				return err
+			}
+			return Print(wd, ctx.App.Stdout)
+		},
+	}
+}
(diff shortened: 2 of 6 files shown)


bmoylan:
Thanks for putting this together so quickly Nick! One comment about error handling but otherwise looks good to me

---

Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[apps/distgo/cmd/projectversion/cmd.go, line 25 at r1](https://reviewable.io:443/reviews/palantir/godel/93#-KgmJC2XOJEylvlRgw1-:-KgmJC2XOJEylvlRgw10:b-uc2jbv) ([raw file](https://github.com/palantir/godel/blob/c55a58a0231fdbb3178a5c1e8ff62066af775a53/apps/distgo/cmd/projectversion/cmd.go#L25)):*
> ```Go
> 	return cli.Command{
> 		Name:  "project-version",
> 		Usage: "Prints the version of the project (basd on Git information)",
> ```

sp: `based`

---

*[apps/distgo/cmd/projectversion/projectversion.go, line 27 at r1](https://reviewable.io:443/reviews/palantir/godel/93#-KgmLAJm19LnoOKRnvza:-KgmLAJm19LnoOKRnvzb:b-z24al2) ([raw file](https://github.com/palantir/godel/blob/c55a58a0231fdbb3178a5c1e8ff62066af775a53/apps/distgo/cmd/projectversion/projectversion.go#L27)):*
> ```Go
> 	projectInfo, err := git.NewProjectInfo(wd)
> 	version := "unspecified"
> 	if err == nil {
> ```

This error gets dropped if it's non-nil, we should probably return it even though we print either way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/93#-:-KgmNkafatAGGcVXpEng:b-lqga6u)*

nmiyake:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[apps/distgo/cmd/projectversion/cmd.go, line 25 at r1](https://reviewable.io:443/reviews/palantir/godel/93#-KgmJC2XOJEylvlRgw1-:-Kgocmh0ZvfxBZDg8pZ_:b1rg8a4) ([raw file](https://github.com/palantir/godel/blob/c55a58a0231fdbb3178a5c1e8ff62066af775a53/apps/distgo/cmd/projectversion/cmd.go#L25)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

sp: `based`
</blockquote></details>

D'oh! Done.

---

*[apps/distgo/cmd/projectversion/projectversion.go, line 27 at r1](https://reviewable.io:443/reviews/palantir/godel/93#-KgmLAJm19LnoOKRnvza:-Kgod4Db0-cxV2yW_NKU:b-ww7o2z) ([raw file](https://github.com/palantir/godel/blob/c55a58a0231fdbb3178a5c1e8ff62066af775a53/apps/distgo/cmd/projectversion/projectversion.go#L27)):*
<details><summary><i>Previously, bmoylan (Brad Moylan) wrote…</i></summary><blockquote>

This error gets dropped if it's non-nil, we should probably return it even though we print either way.
</blockquote></details>

This is the same behavior used by the `build` command (which I believe mirrors the behavior of `gradle-git-version`) -- if any error occurs at the `git` level, it just returns `unspecified`. This can be confusing if a `git` repo has not been initialized or the `git` command does not exist, but it is current behavior.

If we print `unspecified` but still return the error, then I believe the behavior will be that we print `unspecified` but then the command will exit with a non-zero exit code and also print the error message.

Don't feel strongly about this either way -- if we want to error, would prefer to error without printing `unspecified`; otherwise, would prefer to print `unspecified` but not error.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/93)*

bmoylan:
:lgtm:

---

Reviewed 1 of 1 files at r2.

---

*[apps/distgo/cmd/projectversion/projectversion.go, line 27 at r1](https://reviewable.io:443/reviews/palantir/godel/93#-KgmLAJm19LnoOKRnvza:-Kgoq9-00ZiI66fZT1GF:brnmggs) ([raw file](https://github.com/palantir/godel/blob/c55a58a0231fdbb3178a5c1e8ff62066af775a53/apps/distgo/cmd/projectversion/projectversion.go#L27)):*

This is the same behavior used by the `build` command (which I believe mirrors the behavior of `gradle-git-version`) -- if any error occurs at the `git` level, it just returns `unspecified`. This can be confusing if a `git` repo has not been initialized or the `git` command does not exist, but it is current behavior.

If we print `unspecified` but still return the error, then I believe the behavior will be that we print `unspecified` but then the command will exit with a non-zero exit code and also print the error message.

Don't feel strongly about this either way -- if we want to error, would prefer to error without printing `unspecified`; otherwise, would prefer to print `unspecified` but not error.
</blockquote></details>

cool, that sounds fine to me

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/93#-:-KgoqBQw1Ev52RUavnjm:bnfp4nl)*

PR #94, opened 2017-04-03
Dyeduguru/docker dist

yvdinesh:
This change is [Reviewable](https://reviewable.io/reviews/palantir/godel/94)


apps/distgo/cmd/dist/dist.go
```diff
@@ -29,28 +29,35 @@ import (
...
+	if len(specsToBuild) > 0 {
+		if err := build.Run(specsToBuild, nil, build.DefaultContext(), stdout); err != nil {
+			return errors.Wrapf(err, "Failed to build products required for dist")
 		}
-		return cmd.ProcessSerially(Run)(buildSpecWithDeps, stdout)
-	}, cfg, products, wd, stdout)
+	}
+	orderedBuildSpecs, err := Schedule(buildSpecsWithDeps)
+	if err != nil {
+		return errors.Wrapf(err, "")
```
nmiyake: nice thing about the `errors` library is that this is not necessary -- if you want to add context to the error, then fill out the message rather than adding the empty string. If you do not need to add extra content, then just return `err` directly (no need to wrap).

apps/distgo/cmd/dist/dist.go
```diff
@@ -68,99 +75,106 @@ func Run(buildSpecWithDeps params.ProductBuildSpecWithDeps, stdout io.Writer) er
...
-		if currDistCfg.Info.Type() == params.RPMDistType {
-			osArchs := buildSpec.Build.OSArchs
-			expected := osarch.OSArch{OS: "linux", Arch: "amd64"}
-			if len(osArchs) != 1 || osArchs[0] != expected {
-				return fmt.Errorf("RPM is only supported for %v", expected)
-			}
-			if err := checkRPMDependencies(); err != nil {
-				return err
+
+	distTypeOrdering := []params.DistInfoType{params.BinDistType, params.SLSDistType, params.RPMDistType, params.DockerDistType}
```
nmiyake: Hard-coding this logic here seems a bit off. I think that it would be better to have a separate function that orders the `buildspec.Dist` contents according to these rules.

apps/distgo/cmd/dist/dist.go
```diff
@@ -68,99 +75,106 @@ func Run(buildSpecWithDeps params.ProductBuildSpecWithDeps, stdout io.Writer) er
...
+	for _, curDistType := range distTypeOrdering {
```
nmiyake: Other abbreviations of `current` in this file use `curr`, so this should be `currDistType`

apps/distgo/cmd/dist/dockerdist.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/palantir/stacktrace"
+	"github.com/pkg/errors"
+	"os/exec"
+	"path"
+)
+
+func dockerDist(buildSpecWithDeps params.ProductBuildSpecWithDeps, distCfg params.Dist) (Packager, error) {
+	var dockerDistInfo params.DockerDistInfo
+	if info, ok := distCfg.Info.(*params.DockerDistInfo); ok {
+		dockerDistInfo = *info
```
nmiyake: Invert the check to make this block cleaner:

```go
if _, ok := distCfg.Info.(*params.DockerDistInfo); !ok {
    return nil, stacktrace.NewError("Dist info provided is not of type docker info")
}
dockerDistInfo := *distCfg.Info.(*params.DockerDistInfo)
```

apps/distgo/cmd/dist/dockerdist.go
```diff
@@ -0,0 +1,48 @@
...
+
+func buildWithCmd(tag string, contextDir string) error {
+	var args []string
+	args = append(args, "build")
+	args = append(args, "--tag", tag)
+	args = append(args, contextDir)
+
+	dockerBuild := exec.Command("docker", args...)
+	if output, err := dockerBuild.CombinedOutput(); err != nil {
+		fmt.Printf("docker build failed with error:\n%s", string(output))
```
nmiyake: Add `\n` to the end of the ouput.

apps/distgo/cmd/dist/dockerdist.go
```diff
@@ -0,0 +1,48 @@
...
+		return stacktrace.Propagate(err, "failed to run")
```
nmiyake: Use `errors` rather than `stacktrace` and make message more descriptive.

apps/distgo/cmd/dist/rpmdist.go
```diff
@@ -45,6 +46,14 @@ func checkRPMDependencies() error {
 
 func rpmDist(buildSpecWithDeps params.ProductBuildSpecWithDeps, distCfg params.Dist, outputProductDir string, stdout io.Writer) (p Packager, rErr error) {
 	buildSpec := buildSpecWithDeps.Spec
+	osArchs := buildSpec.Build.OSArchs
+	expected := osarch.OSArch{OS: "linux", Arch: "amd64"}
+	if len(osArchs) != 1 || osArchs[0] != expected {
+		return nil, fmt.Errorf("RPM is only supported for %v", expected)
+	}
+	if err := checkRPMDependencies(); err != nil {
```
nmiyake: Why is this being re-added? Seems unrelated to the commit. You may have based your work on an old commit?

apps/distgo/cmd/dist/scheduler.go
```diff
@@ -0,0 +1,93 @@
+package dist
+
+import (
+	"github.com/palantir/godel/apps/distgo/params"
+	"github.com/pkg/errors"
+	"sort"
```
nmiyake: Doesn't look like these imports are sorted.

apps/distgo/cmd/dist/scheduler.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"sort"
+
+	"github.com/pkg/errors"
+
+	"github.com/palantir/godel/apps/distgo/params"
+)
+
+func Schedule(specsWithDeps []params.ProductBuildSpecWithDeps) ([]params.ProductBuildSpecWithDeps, error) {
```
nmiyake: `Schedule` isn't the most descriptive name here -- when I see `dist.Schedule`, I would expect such a function to actually schedule actions rather than just ordering input.

Maybe something like `OrderSpecs` or `OrderBuildSpecs` or `SortBuildSpecsTopologically`? Comment describing what the function does would also be helpful.

apps/distgo/cmd/dist/scheduler.go
```diff
@@ -0,0 +1,109 @@
...
+	var schedule []params.ProductBuildSpecWithDeps
+	graph := make(map[string]map[string]bool)
+	specMap := make(map[string][]params.ProductBuildSpecWithDeps)
+	// create a graph of dependencies
+	for _, curSpec := range specsWithDeps {
+		product := curSpec.Spec.ProductName
+		if graph[product] == nil {
+			graph[product] = make(map[string]bool)
+		}
+		if specMap[product] == nil {
```
nmiyake: No need for this `if` block -- although a map must be non-`nil` to add/assign entries, `nil` slices are fine (you can still append to them).

apps/distgo/cmd/dist/scheduler.go
```diff
@@ -0,0 +1,109 @@
...
+			specMap[product] = make([]params.ProductBuildSpecWithDeps, 0)
+		}
+		specMap[product] = append(specMap[product], curSpec)
+		for _, curDist := range curSpec.Spec.Dist {
+			for _, dep := range curDist.Info.Deps() {
+				if graph[dep] == nil {
+					graph[dep] = make(map[string]bool)
+				}
+				graph[dep][product] = true
```
nmiyake: Looks like `graph[dep][product]` is only ever set to `true`, so internal map is effectively a set. In that case, more semantically clear to declare inner map as `map[string]struct{}` and to do assignment as `graph[dep][product] = struct{}{}`

apps/distgo/cmd/dist/scheduler.go
```diff
@@ -0,0 +1,109 @@
...
+
+func Schedule(specsWithDeps []params.ProductBuildSpecWithDeps) ([]params.ProductBuildSpecWithDeps, error) {
+	var schedule []params.ProductBuildSpecWithDeps
+	graph := make(map[string]map[string]bool)
+	specMap := make(map[string][]params.ProductBuildSpecWithDeps)
+	// create a graph of dependencies
+	for _, curSpec := range specsWithDeps {
+		product := curSpec.Spec.ProductName
+		if graph[product] == nil {
+			graph[product] = make(map[string]bool)
```
nmiyake: Don't think this is actually necessary because you never actually assign into `graph[product]` at this point (and you properly initialize `graph[dep]` before assigning into `graph[dep][product]`)

apps/distgo/cmd/dist/scheduler.go
```diff
@@ -0,0 +1,109 @@
...
+	var nodes []string
+	indeg := make(map[string]int)
+	for node := range graph {
+		indeg[node] = 0
+		nodes = append(nodes, node)
+	}
+	sort.Strings(nodes)
+	// compute the incoming edges on each vertex
+	for _, v := range nodes {
+		for neighbour := range graph[v] {
```
nmiyake: Would like to stick with American English in this project, so prefer `neighbor` :)

apps/distgo/cmd/dist/scheduler.go
```diff
@@ -0,0 +1,109 @@
...
+		sort.Strings(neighbours)
+		for _, neighbour := range neighbours {
+			indeg[neighbour]--
+			if indeg[neighbour] == 0 {
+				q = append(q, neighbour)
+			}
+		}
+	}
+	if len(order) != len(graph) {
+		return nil, errors.New("Please provide a valid DAG as input")
```
nmiyake: Is there enough information to provide context on what was illegal about the DAG? (what causes a cycle, etc.)?

apps/distgo/cmd/dist/scheduler_test.go
```diff
@@ -0,0 +1,111 @@
...
+			require.Error(t, err)
+			continue
+
+		}
+		require.NoError(t, err)
+		for i, expectedSpec := range testcase.expected {
+			require.Equal(t, expectedSpec.Spec.ProductName, actual[i].Spec.ProductName)
+		}
+	}
+
```
nmiyake: Remove extra newline

apps/distgo/params/dist.go
```diff
@@ -136,3 +148,29 @@ type RPMDistInfo struct {
...
+func (d *DockerDistInfo) Type() DistInfoType {
+	return DockerDistType
+}
+
+func (d *DockerDistInfo) Deps() []string {
+	var deps []string
+	for product := range d.DistDeps {
+		deps = append(deps, product)
+	}
+	return deps
```
nmiyake: Note that, because this iterates over map contents, the return order here will be random between calls. This may be fine, but if you want this to be deterministic, you can do so by adding a `sort.Strings(deps)` before the return to ensure that the items are always sorted lexicographically.

apps/distgo/cmd/dist/dist_test.go
```diff
@@ -1179,3 +1186,169 @@ daemon: true
...
+				), nil)
+				require.NoError(t, err)
+				return []params.ProductBuildSpecWithDeps{fooSpecWithDeps, barSpecWithDeps}
+			},
+			preDistAction: func(projectDir string, buildSpec []params.ProductBuildSpecWithDeps) {
+				gittest.CreateGitTag(t, projectDir, "0.1.0")
+			},
+			validate: func(caseNum int, name string, projectDir string, cli *dockercli.Client) {
+				filter := filters.NewArgs()
+				filter.Add("reference", "foo:0.1.0")
```
nmiyake: If this test is adding Docker images, probably best to make sure the name is more unique (and different between runs) to make it effective. Maybe something like `testDockerDist-foo-{#}` where `{#}` is some form or randomness (current time or something derived from it)?

apps/distgo/cmd/dist/dist_test.go
```diff
@@ -1179,3 +1186,169 @@ daemon: true
...
+				images, err := cli.ImageList(context.Background(), types.ImageListOptions{Filters: filter})
+				require.NoError(t, err)
+				require.True(t, len(images) > 0)
+				filter = filters.NewArgs()
+				filter.Add("reference", "bar:0.1.0")
+				images, err = cli.ImageList(context.Background(), types.ImageListOptions{Filters: filter})
+				require.NoError(t, err)
+				require.True(t, len(images) > 0)
```
nmiyake: If we make the above change this will be slightly less of an issue, but I think it would still be nice to try to remove/clean the images created by the test at the end of the test (basically defer a call to remove the tags after they are created). Generally, tests should strive to leave the host system in the same state as before they were run.

circle.yml
```diff
@@ -6,6 +6,8 @@ stages:
     workDir: ~/go/src/github.com/palantir/godel
     machine:
       reusable: true
+      services:
+        - docker
```
nmiyake: This is necessary but probably not quite sufficient -- I believe we have tests/integration tests that run in Docker containers, so those tests will probably not be able to run the Docker dist tests.

We may need to add logic to the tests that test for the existence of the `docker` command and then skip the tests if the command is not present -- presumably, running directly in CircleCI will be fine, but tests that run inside Docker containers will not be able to run the Docker dist tests.

apps/distgo/cmd/dist/dist.go
```diff
@@ -29,28 +29,35 @@ import (
...
-			if forceBuild {
-				specsToBuild = append(specsToBuild, currSpecWithDeps.AllSpecs()...)
-			} else {
-				specsToBuild = append(specsToBuild, build.RequiresBuild(currSpecWithDeps, nil).Specs()...)
-			}
+	buildSpecsWithDeps, err := build.SpecsWithDepsForArgs(cfg, products, wd)
+	if err != nil {
+		return errors.Wrapf(err, "failed to construct the spec from config")
+	}
+	// build the
```
yvdinesh: completed the comment

apps/distgo/cmd/dist/dist.go
yvdinesh: done.

apps/distgo/cmd/dist/dist.go
yvdinesh: done.

apps/distgo/cmd/dist/dist.go
```diff
@@ -68,99 +75,106 @@ func Run(buildSpecWithDeps params.ProductBuildSpecWithDeps, stdout io.Writer) er
...
+		// Order of executing dist types for each product:
+		// bin -> sls -> rpm -> docker
+		for _, currDistCfg := range buildSpec.Dist {
+			if currDistCfg.Info.Type() != curDistType {
+				continue
 			}
-		}
 
-		outputDir := path.Join(buildSpec.ProjectDir, currDistCfg.OutputDir)
+			if currDistCfg.Info.Type() == params.DockerDistType {
```
yvdinesh: removed.

apps/distgo/cmd/dist/scheduler_test.go
```diff
@@ -0,0 +1,95 @@
...
+						DistDeps: deps,
+					},
+				},
+			},
+		},
+	}
+}
+
+func TestSchedule(t *testing.T) {
+
```
yvdinesh: done.

apps/distgo/cmd/dist/scheduler_test.go
yvdinesh: done.

apps/distgo/params/dist.go
```diff
@@ -22,7 +22,7 @@ type Dist struct {
 	// OutputDir is the directory to which the distribution is written.
 	OutputDir string
 
-	// InputDir is the path (from the project root) to a directory whose contents will be copied into the output
+	// ContextDir is the path (from the project root) to a directory whose contents will be copied into the output
```
yvdinesh: argh, gogland refactors comments as well. will change it back.

apps/distgo/params/dist.go
```diff
@@ -83,6 +85,11 @@ func (i *BinDistInfo) Type() DistInfoType {
 	return BinDistType
 }
 
+func (i *BinDistInfo) Deps() []string {
+	// no deps for bin type
+	return []string{}
```
yvdinesh: done.

apps/distgo/params/dist.go
```diff
@@ -118,6 +125,11 @@ func (i *SLSDistInfo) Type() DistInfoType {
 	return SLSDistType
 }
 
+func (i *SLSDistInfo) Deps() []string {
+	// no deps for sls type
+	return []string{}
```
yvdinesh: done.

apps/distgo/params/dist.go
```diff
@@ -136,3 +148,29 @@ type RPMDistInfo struct {
 func (i *RPMDistInfo) Type() DistInfoType {
 	return RPMDistType
 }
+
+func (i *RPMDistInfo) Deps() []string {
+	// no deps for rpm type
+	return []string{}
```
yvdinesh: done.

apps/distgo/params/dist.go
yvdinesh: added sorting.

vendor/github.com/Microsoft/go-winio/README.md
```diff
@@ -0,0 +1,22 @@
+# go-winio
```
yvdinesh: done.

vendor/github.com/Sirupsen/logrus/CHANGELOG.md
```diff
@@ -0,0 +1,94 @@
+# 0.11.5
```
yvdinesh: done.

vendor/github.com/Sirupsen/logrus/README.md
```diff
@@ -0,0 +1,479 @@
+# Logrus <img src="http://i.imgur.com/hTeVwmJ.png" width="40" height="40" alt=":walrus:" class="emoji" title=":walrus:"/>&nbsp;[![Build Status](https://travis-ci.org/Sirupsen/logrus.svg?branch=master)](https://travis-ci.org/Sirupsen/logrus)&nbsp;[![GoDoc](https://godoc.org/github.com/Sirupsen/logrus?status.svg)](https://godoc.org/github.com/Sirupsen/logrus)
```
yvdinesh: done.

vendor/github.com/docker/docker/api/swagger.yaml
```diff
@@ -0,0 +1,8134 @@
+# A Swagger 2.0 (a.k.a. OpenAPI) definition of the Engine API.
```
yvdinesh: done.

vendor/github.com/docker/docker/api/swagger-gen.yaml
```diff
@@ -0,0 +1,12 @@
+
```
yvdinesh: done.

vendor/github.com/docker/docker/api/README.md
```diff
@@ -0,0 +1,42 @@
+# Working on the Engine API
```
yvdinesh: done.

vendor/github.com/docker/go-units/README.md
```diff
@@ -0,0 +1,16 @@
+[![GoDoc](https://godoc.org/github.com/docker/go-units?status.svg)](https://godoc.org/github.com/docker/go-units)
```
yvdinesh: done.

vendor/github.com/docker/go-units/MAINTAINERS
```diff
@@ -0,0 +1,27 @@
+# go-connections maintainers file
```
yvdinesh: done.

vendor/github.com/docker/go-units/CONTRIBUTING.md
```diff
@@ -0,0 +1,67 @@
+# Contributing to go-units
```
yvdinesh: done.

vendor/github.com/docker/go-units/circle.yml
```diff
@@ -0,0 +1,11 @@
+dependencies:
```
yvdinesh: done.

vendor/github.com/docker/libtrust/CONTRIBUTING.md
```diff
@@ -0,0 +1,13 @@
+# Contributing to libtrust
```
yvdinesh: done.

vendor/github.com/docker/libtrust/MAINTAINERS
```diff
@@ -0,0 +1,3 @@
+Solomon Hykes <solomon@docker.com>
```
yvdinesh: done.

vendor/github.com/docker/libtrust/README.md
```diff
@@ -0,0 +1,22 @@
+# libtrust
```
yvdinesh: done.

vendor/github.com/opencontainers/go-digest/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
+# Contributing to Docker open source projects
```
yvdinesh: done.

vendor/github.com/opencontainers/go-digest/MAINTAINERS
```diff
@@ -0,0 +1,9 @@
+Aaron Lehmann <aaron.lehmann@docker.com> (@aaronlehmann)
```
yvdinesh: done.

vendor/github.com/opencontainers/go-digest/README.md
```diff
@@ -0,0 +1,104 @@
+# go-digest
```
yvdinesh: done.

apps/distgo/cmd/dist/dockerdist.go
yvdinesh: done.

apps/distgo/cmd/dist/dockerdist.go
yvdinesh: done.

apps/distgo/cmd/dist/rpmdist.go
yvdinesh: I moved this from dist.go to rpmdist.go. This check should be performed by rpmdist since its specific to it.

apps/distgo/cmd/dist/scheduler.go
yvdinesh: done.

apps/distgo/cmd/dist/scheduler.go
yvdinesh: done.

apps/distgo/cmd/dist/scheduler.go
yvdinesh: done.

apps/distgo/cmd/dist/scheduler.go
yvdinesh: done.

apps/distgo/cmd/dist/scheduler.go
yvdinesh: done.

apps/distgo/cmd/dist/scheduler.go
yvdinesh: Since the given graph is directed, leftover vertices with non zero in-degrees means that it contains a cycle. I will make the error indicate that the dependencies form a cycle.

apps/distgo/cmd/dist/scheduler.go
yvdinesh: we actually need this step because this creates a vertex in the graph even if there are no edges in the graph and this would result in the ordering to be of same length as the number of vertices in the input graph.

nmiyake: OK that makes sense

apps/distgo/cmd/dist/dist_test.go
yvdinesh: added random padding to images.

apps/distgo/cmd/dist/dist_test.go
yvdinesh: added random padding to images and changed the name to have `test-docker-dist` prefix.

circle.yml
yvdinesh: synced with @nmiyake. Decided on adding a new test tag for the distgo tests which require docker executable.

apps/distgo/cmd/dist/dist.go
yvdinesh: done.

yvdinesh:
@nmiyake ready for review.

nmiyake:
Much better! A few minor comments. Will take a look at adding exclusions for test matchers as well since I really don't like adding the `apps/docker` directory.

---

Reviewed 518 of 546 files at r1, 13 of 32 files at r2.
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 49 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-KgyvYqVFRlp2Ot9B2I8:-KgyvYqX0A192GW-C-fJ:bd8xiwl) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/dockerdist.go#L49)):*
> ```Go
> }
> 
> func buildWithCmd(tag string, contextDir string) error {
> ```

For consecutive parameters with same type, use form `(tag, contextDir string)`

---

*[apps/distgo/cmd/dist/dockerdist.go, line 57 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgyvlspr1SShfFXPtED:-KgyvlsqeOjnmnL1p1Cs:b-egt5yb) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/dockerdist.go#L57)):*
> ```Go
> 	dockerBuild := exec.Command("docker", args...)
> 	if output, err := dockerBuild.CombinedOutput(); err != nil {
> 		fmt.Printf("docker build failed with error:\n%s\n", string(output))
> ```

Avoid `fmt.Printf` -- either include this information as part of the returned error or print it to the output stream of the program (in which case you'll need to pass through an `io.Writer` as an argument)

---

*[apps/distgo/cmd/dist/order_dist.go, line 22 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-KgywgTKd36jdzq9Ko23:-KgywgTKd36jdzq9Ko24:b-s06a9t) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_dist.go#L22)):*
> ```Go
> 
> var (
> 	distTypeOrdering = []params.DistInfoType{params.BinDistType,
> ```

When declaring something like this, keep one item per line so it's easier to see all of them. Also, no need to do a block declaration when there's only a single variable.

```go
var distTypeOrdering = []params.DistInfoType{
    params.BinDistType,
    params.SLSDistType,
    params.RPMDistType,
    params.DockerDistType,
}
```

---

*[apps/distgo/cmd/dist/order_dist.go, line 31 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-KgyxFFJ3C1XwXkVuNMx:-KgyxFFJ3C1XwXkVuNMy:bc8upxs) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_dist.go#L31)):*
> ```Go
> 	var orderedDists []params.Dist
> 	for _, currDistType := range distTypeOrdering {
> 		for _, currDist := range dists {
> ```

This is an n^2 double for loop for sorting. In this case the perf isn't much (because the inputs will always be relatively small) so that's not a huge issue, but it's not as easy to follow conceptually either.

The better general approach would be to implement this as a sort. You can use the `sort.Slice` function (https://golang.org/pkg/sort/#Slice, new in Go 1.8) to do this without having to declare an entirely new type. Would look something like:

```go
var orderedDists []parms.Dist
for _, v := range dists {
     orderedDists = append(orderedDists, v)
}
sort.Slice(orderedDists, func(i, j int) bool { return orderedDists[i].Info.Type() < orderedDists[j].Info.Type() })
```

---

*[apps/distgo/cmd/dist/order_specs_test.go, line 100 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-AM0zm3ZMoR9XeJp:-Kgz-AM0zm3ZMoR9XeJq:bbl9t0t) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_specs_test.go#L100)):*
> ```Go
> 		actual, err := dist.OrderBuildSpecs(testcase.input)
> 		if !testcase.valid {
> 			require.Error(t, err)
> ```

You may or may not want to separate out testing for errors into a different test function, but either way would generally suggest that, when testing errors, you have something like and `expectErr` that you compare against to make sure that the error message is what you expect (can either be an exact string or a regular expression)

---

*[apps/distgo/cmd/dist/order_specs_test.go, line 102 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-5sYBPulikFD5W_B:-Kgz-5sYBPulikFD5W_C:bsriq5) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_specs_test.go#L102)):*
> ```Go
> 			require.Error(t, err)
> 			continue
> ```

Remove extra newline

---

*[apps/distgo/cmd/dist/order_specs_test.go, line 106 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-KgyzsLiczQwNi464Z8d:-KgyzsLiczQwNi464Z8e:b-bif7jf) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_specs_test.go#L106)):*
> ```Go
> 		require.NoError(t, err)
> 		for i, expectedSpec := range testcase.expected {
> 			require.Equal(t, expectedSpec.Spec.ProductName, actual[i].Spec.ProductName)
> ```

When you're doing the actual test/comparison, use `assert.Equal` instead (so that even if one case fails, the other cases will still be run). We use `require` in setup steps because typically the test doesn't make sense to run if there is an error in one of the setup steps.

---

*[apps/distgo/cmd/dist/rpmdist.go, line 52 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-TWC2gmnqGVtpWfR:-Kgz-TWC2gmnqGVtpWfS:b-5smdw5) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/rpmdist.go#L52)):*
> ```Go
> 	expected := osarch.OSArch{OS: "linux", Arch: "amd64"}
> 	if len(osArchs) != 1 || osArchs[0] != expected {
> 		return nil, fmt.Errorf("RPM is only supported for %v", expected)
> ```

Why are you adding this? is this code moved from somewhere else, or are you adding it as new logic?

Pretty sure that we explicitly don't want to do this -- I think that, if a `darwin` (or other system) has the `rpm` command installed, then they should be allowed to run `rpm` dist tasks.

---

*[apps/distgo/params/dist.go, line 159 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-uocKG6OWOa6gMx2:-Kgz-uodppvRuAkttR05:b-dm9xo6) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/params/dist.go#L159)):*
> ```Go
> }
> 
> type DockerDeps map[string][]DistInfoType
> ```

Since we don't have any functions defined on this alias, not totally sure if it's worth declaring an alias specifically for this type.

---

*[apps/docker/distgo/dist_docker_test.go, line 64 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz1leNXOcMnSqtd1E4:-Kgz1leNXOcMnSqtd1E5:b-mwa4y0) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/docker/distgo/dist_docker_test.go#L64)):*
> ```Go
> 	for i, currCase := range []struct {
> 		name          string
> 		skip          func() bool
> ```

Hmm, don't see any cases actually define this -- if it's not used, we should just remove.

---

*[apps/docker/distgo/dist_docker_test.go, line 67 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz1q_IjJhwSzsYUWc5:-Kgz1q_IjJhwSzsYUWc6:bb73xfv) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/docker/distgo/dist_docker_test.go#L67)):*
> ```Go
> 		spec          func(projectDir string, randomPad string) []params.ProductBuildSpecWithDeps
> 		preDistAction func(projectDir string, buildSpec []params.ProductBuildSpecWithDeps)
> 		skipBuild     bool
> ```

Same

---

*[godel/config/test.yml, line 7 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz21AJNiWA3Qvy3zLX:-Kgz21AJNiWA3Qvy3zLY:b-i1pofs) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/godel/config/test.yml#L7)):*
> ```YAML
>   docker-distgo:
>     paths:
>       - "apps/docker/distgo"
> ```

Ugh really don't like this -- `apps` is meant to be a top-level directory with applications, so this would indicate that `docker` is an application.

I'm fine keeping this while we iterate on this PR, but my preference would be to introduce another PR/feature first that does the test tags exclusions that we actually want/need here and use that instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/94#-:-Kgz91CZJs2ZIQeFW1H1:b-q39lm0)*

yvdinesh:
Review status: all files reviewed at latest revision, 12 unresolved discussions.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 49 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-KgyvYqVFRlp2Ot9B2I8:-KgzCcwwGo8rPcIulZC4:b-896fix) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/dockerdist.go#L49)):*

For consecutive parameters with same type, use form `(tag, contextDir string)`
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 57 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgyvlspr1SShfFXPtED:-KgzCxbixbA6bi-fEtkX:b-1qtf6d) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/dockerdist.go#L57)):*

Avoid `fmt.Printf` -- either include this information as part of the returned error or print it to the output stream of the program (in which case you'll need to pass through an `io.Writer` as an argument)
</blockquote></details>

added the information to the returned error.

---

*[apps/distgo/cmd/dist/order_dist.go, line 22 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-KgywgTKd36jdzq9Ko23:-KgzE4Q2oonjYBIbBa_Q:b-896fix) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_dist.go#L22)):*

When declaring something like this, keep one item per line so it's easier to see all of them. Also, no need to do a block declaration when there's only a single variable.

```go
var distTypeOrdering = []params.DistInfoType{
    params.BinDistType,
    params.SLSDistType,
    params.RPMDistType,
    params.DockerDistType,
}
```
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/order_dist.go, line 31 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-KgyxFFJ3C1XwXkVuNMx:-KgzHFfI0A7ueB084MZv:b-896fix) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_dist.go#L31)):*

This is an n^2 double for loop for sorting. In this case the perf isn't much (because the inputs will always be relatively small) so that's not a huge issue, but it's not as easy to follow conceptually either.

The better general approach would be to implement this as a sort. You can use the `sort.Slice` function (https://golang.org/pkg/sort/#Slice, new in Go 1.8) to do this without having to declare an entirely new type. Would look something like:

```go
var orderedDists []parms.Dist
for _, v := range dists {
     orderedDists = append(orderedDists, v)
}
sort.Slice(orderedDists, func(i, j int) bool { return orderedDists[i].Info.Type() < orderedDists[j].Info.Type() })
```
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/order_specs_test.go, line 100 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-AM0zm3ZMoR9XeJp:-KgzOVseKbrz9LNZ-cLb:b-896fix) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_specs_test.go#L100)):*

You may or may not want to separate out testing for errors into a different test function, but either way would generally suggest that, when testing errors, you have something like and `expectErr` that you compare against to make sure that the error message is what you expect (can either be an exact string or a regular expression)
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/order_specs_test.go, line 102 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-5sYBPulikFD5W_B:-KgzL8PLUh-cqFax_5k5:b-896fix) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_specs_test.go#L102)):*

Remove extra newline
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/order_specs_test.go, line 106 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-KgyzsLiczQwNi464Z8d:-KgzLExlwqPFaEha7i9m:b-896fix) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/order_specs_test.go#L106)):*

When you're doing the actual test/comparison, use `assert.Equal` instead (so that even if one case fails, the other cases will still be run). We use `require` in setup steps because typically the test doesn't make sense to run if there is an error in one of the setup steps.
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/rpmdist.go, line 52 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-TWC2gmnqGVtpWfR:-KgzLGPvOwoe7TtXz6Jj:b-qxejaw) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/rpmdist.go#L52)):*

Why are you adding this? is this code moved from somewhere else, or are you adding it as new logic?

Pretty sure that we explicitly don't want to do this -- I think that, if a `darwin` (or other system) has the `rpm` command installed, then they should be allowed to run `rpm` dist tasks.
</blockquote></details>

I am not adding this. this is already existing code which was part of dist.go which was doing validation for rpm dist type. So i moved it to rpmdist.go, as it is more relevant there.

---

*[apps/distgo/params/dist.go, line 159 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-uocKG6OWOa6gMx2:-KgzN4qy4t2q885vGMnO:b5quwf4) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/params/dist.go#L159)):*

Since we don't have any functions defined on this alias, not totally sure if it's worth declaring an alias specifically for this type.
</blockquote></details>

removed the type.

---

*[apps/docker/distgo/dist_docker_test.go, line 64 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz1leNXOcMnSqtd1E4:-KgzLqGcelAmN0GP7Uw-:b-896fix) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/docker/distgo/dist_docker_test.go#L64)):*

Hmm, don't see any cases actually define this -- if it's not used, we should just remove.
</blockquote></details>

Done.

---

*[apps/docker/distgo/dist_docker_test.go, line 67 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz1q_IjJhwSzsYUWc5:-KgzM-1iCVXTtKY1KIhl:b-896fix) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/docker/distgo/dist_docker_test.go#L67)):*

Same
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/94)*

nmiyake:
Looks good -- a few minor things, and then request to rebase on latest gödel and use the tag exclude config so we can put the Docker dist tests back under `distgo`

---

Reviewed 1 of 32 files at r2, 5 of 6 files at r3.
Review status: 528 of 529 files reviewed at latest revision, 3 unresolved discussions.

---

*[apps/distgo/cmd/dist/order_dist_test.go, line 58 at r3](https://reviewable.io:443/reviews/palantir/godel/94#-Kh0QjrOPOrzqt5xK6Sl:-Kh0QjrOPOrzqt5xK6Sm:bobkpfo) ([raw file](https://github.com/palantir/godel/blob/42b51526fb81ad0af6850f6c440b2c607e7d6987/apps/distgo/cmd/dist/order_dist_test.go#L58)):*
> ```Go
> 		actual := dist.OrderProductDists(testCase.input)
> 		for i, d := range actual {
> 			require.True(t, d.Info.Type() == testCase.expected[i].Info.Type(), "Case: %v Index: %v", caseNum, i)
> ```

Minor, but this should be `assert.Equal` (`assert.` because it's the assertion/test rather than a setup step, and `Equal` because that is what is being checked here).

---

*[apps/distgo/cmd/dist/rpmdist.go, line 52 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-TWC2gmnqGVtpWfR:-Kh0RCH9qSDoPQGMlEL3:b-ftpfys) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/rpmdist.go#L52)):*

I am not adding this. this is already existing code which was part of dist.go which was doing validation for rpm dist type. So i moved it to rpmdist.go, as it is more relevant there.
</blockquote></details>

OK sorry I remember you saying this in a previous review as well. If that's the case, why don't I see the removal of this logic as part of the same PR?

---

*[godel/config/test.yml, line 7 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz21AJNiWA3Qvy3zLX:-Kh0RQt4swiqvIOrd_hK:b-wqtvfe) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/godel/config/test.yml#L7)):*

Ugh really don't like this -- `apps` is meant to be a top-level directory with applications, so this would indicate that `docker` is an application.

I'm fine keeping this while we iterate on this PR, but my preference would be to introduce another PR/feature first that does the test tags exclusions that we actually want/need here and use that instead.
</blockquote></details>

Merged change in gödel that adds `Exclude` for tags, so if you rebase on master you should be able to do this correctly now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/94#-:-Kh0RZSWff1dobDMCvHs:b-yjf7c0)*

nmiyake:
Reviewed 1 of 6 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/94)*

yvdinesh:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[apps/distgo/cmd/dist/order_dist_test.go, line 58 at r3](https://reviewable.io:443/reviews/palantir/godel/94#-Kh0QjrOPOrzqt5xK6Sl:-Kh3NRqQ14c-pzJy60F2:b-896fix) ([raw file](https://github.com/palantir/godel/blob/42b51526fb81ad0af6850f6c440b2c607e7d6987/apps/distgo/cmd/dist/order_dist_test.go#L58)):*

Minor, but this should be `assert.Equal` (`assert.` because it's the assertion/test rather than a setup step, and `Equal` because that is what is being checked here).
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/rpmdist.go, line 52 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-TWC2gmnqGVtpWfR:-Kh3NV_OBherKE81Vy5o:b-c8jcbk) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/rpmdist.go#L52)):*

OK sorry I remember you saying this in a previous review as well. If that's the case, why don't I see the removal of this logic as part of the same PR?
</blockquote></details>

This is moved from a different file. It was originally in `dist.go` and i moved this to `rpmdist.go`

---

*[godel/config/test.yml, line 7 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz21AJNiWA3Qvy3zLX:-Kh3NedP06pRLAAI8hEr:b-896fix) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/godel/config/test.yml#L7)):*

Merged change in gödel that adds `Exclude` for tags, so if you rebase on master you should be able to do this correctly now.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/94)*

nmiyake:
Thanks -- minor feedback on content of exclude clause in config, but think this should be pretty much ready after that.

---

Reviewed 7 of 7 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[godel/config/test.yml, line 7 at r4](https://reviewable.io:443/reviews/palantir/godel/94#-Kh3pKqRBAybSELQxWjz:-Kh3pKqRBAybSELQxWk-:b-famlsg) ([raw file](https://github.com/palantir/godel/blob/ba44df1558623dd7965612c3b14b07a40cc19cd7/godel/config/test.yml#L7)):*
> ```YAML
>     exclude:
>       names:
>         - "dockertests"
> ```

I think it makes more sense for the `exclude` block to be:

```
exclude:
  paths:
    - "apps/distgo/dockertests"
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/94#-:-Kh3pYdWAClVQVYgHDCc:b-zgu4f)*

nmiyake:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[apps/distgo/cmd/dist/rpmdist.go, line 52 at r2](https://reviewable.io:443/reviews/palantir/godel/94#-Kgz-TWC2gmnqGVtpWfR:-Kh3pzDechTn7Ie_Q6dF:bhd7ddr) ([raw file](https://github.com/palantir/godel/blob/7a533c409bf86e0f894012201b4dfdaf1ecca8b7/apps/distgo/cmd/dist/rpmdist.go#L52)):*

This is moved from a different file. It was originally in `dist.go` and i moved this to `rpmdist.go`
</blockquote></details>

OK yup see it now -- thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/94)*

yvdinesh:

---

*[godel/config/test.yml, line 7 at r4](https://reviewable.io:443/reviews/palantir/godel/94#-Kh3pKqRBAybSELQxWjz:-Kh3qZA1EJBv16X_e04Y:b-896fix) ([raw file](https://github.com/palantir/godel/blob/ba44df1558623dd7965612c3b14b07a40cc19cd7/godel/config/test.yml#L7)):*

I think it makes more sense for the `exclude` block to be:

```
exclude:
  paths:
    - "apps/distgo/dockertests"
```
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/94)*

nmiyake:
:lgtm:

Would like to play with this a bit before we cut the next gödel release, but I think this is ready to merge so that we can do so. Thanks for iterating on this!

---

Reviewed 1 of 1 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/94#-:-Kh3wFlAEg-iJIM3vnRp:b4mf6jl)*

PR #97, opened 2017-04-07
Copy dist deps to docker context dir in distgo

yvdinesh:
This change is [Reviewable](https://reviewable.io/reviews/palantir/godel/97)

### DIFF
--- a/apps/distgo/cmd/dist/dist.go
+++ b/apps/distgo/cmd/dist/dist.go
@@ -182,7 +182,7 @@ func copyBuildArtifactsToBinDir(buildSpecWithDeps params.ProductBuildSpecWithDep
 	}
 
 	// copy build artifacts for dependent products
-	for _, currDepSpec := range buildSpecWithDeps.Deps {
+	for _, currDepSpec := range buildSpecWithDeps.BuildDeps {
 		if err := copyBuildArtifacts(currDepSpec, binSpecDir); err != nil {
 			return errors.Wrapf(err, "failed to copy build artifacts for %v", currDepSpec.ProductName)
 		}
--- a/apps/distgo/cmd/dist/dockerdist.go
+++ b/apps/distgo/cmd/dist/dockerdist.go
@@ -16,6 +16,7 @@ package dist
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"path"
 
@@ -37,6 +38,31 @@ func dockerDist(buildSpecWithDeps params.ProductBuildSpecWithDeps, distCfg param
 	}
 	completeTag := fmt.Sprintf("%s:%s", dockerDistInfo.Repository, dockerDistInfo.Tag)
 	contextDir := path.Join(buildSpecWithDeps.Spec.ProjectDir, dockerDistInfo.ContextDir)
+
+	// link dependent artifacts into the context directory
+	dockerDeps := dockerDistInfo.DistDeps
+	for depProduct := range dockerDeps {
+		for depDistType := range dockerDeps[depProduct] {
+			if depDistType == params.DockerDistType {
+				// no need to copy docker artifacts
+				continue
+			}
+			depProductSpec := buildSpecWithDeps.DistDeps[depProduct]
+			for _, depDist := range depProductSpec.Dist {
+				if depDist.Info.Type() != depDistType {
+					continue
+				}
+				artifactLocation := ArtifactPath(depProductSpec, depDist)
+				outputLink := dockerDeps[depProduct][depDistType]
+				if outputLink == "" {
+					outputLink = path.Base(artifactLocation)
+				}
+				if err := os.Link(artifactLocation, path.Join(contextDir, outputLink)); err != nil {
+					return nil, err
+				}
+			}
+		}
+	}
 
 	return packager(func() error {
 		if err := buildWithCmd(completeTag, contextDir); err != nil {
--- a/apps/distgo/cmd/dist/dockerdist.go
+++ b/apps/distgo/cmd/dist/dockerdist.go
@@ -40,24 +40,23 @@ func dockerDist(buildSpecWithDeps params.ProductBuildSpecWithDeps, distCfg param
 	contextDir := path.Join(buildSpecWithDeps.Spec.ProjectDir, dockerDistInfo.ContextDir)
 
 	// link dependent artifacts into the context directory
-	dockerDeps := dockerDistInfo.DistDeps
-	for depProduct := range dockerDeps {
-		for depDistType := range dockerDeps[depProduct] {
-			if depDistType == params.DockerDistType {
+	for depProduct, distTypes := range dockerDistInfo.DistDeps.ToMap() {
+		for distType := range distTypes {
+			if distType == params.DockerDistType {
 				// no need to copy docker artifacts
 				continue
 			}
 			depProductSpec := buildSpecWithDeps.DistDeps[depProduct]
 			for _, depDist := range depProductSpec.Dist {
-				if depDist.Info.Type() != depDistType {
+				if depDist.Info.Type() != distType {
 					continue
 				}
 				artifactLocation := ArtifactPath(depProductSpec, depDist)
-				outputLink := dockerDeps[depProduct][depDistType]
-				if outputLink == "" {
-					outputLink = path.Base(artifactLocation)
+				targetFile := distTypes[distType]
+				if targetFile == "" {
+					targetFile = path.Base(artifactLocation)
 				}
-				if err := os.Link(artifactLocation, path.Join(contextDir, outputLink)); err != nil {
+				if err := os.Link(artifactLocation, path.Join(contextDir, targetFile)); err != nil {
 					return nil, err
 				}
 			}
--- a/apps/distgo/cmd/dist/dockertests/dist_docker_test.go
+++ b/apps/distgo/cmd/dist/dockertests/dist_docker_test.go
@@ -171,9 +171,7 @@ func TestDockerDist(t *testing.T) {
 									Tag:        "0.1.0",
 									ContextDir: "foo/docker",
 									DistDeps: params.DockerDistDeps{
-										"bar": {
-											params.DockerDistType: "",
-										},
+										{Product: "bar", DistType: params.DockerDistType, TargetFile: ""},
 									},
 								},
 							},
@@ -303,13 +301,9 @@ func TestDockerDist(t *testing.T) {
 									Tag:        "0.1.0",
 									ContextDir: "bar/docker",
 									DistDeps: params.DockerDistDeps{
-										"bar": {
-											params.SLSDistType: "bar-sls.tgz",
-										},
-										"foo": {
-											params.SLSDistType: "foo-sls.tgz",
-											params.BinDistType: "foo-bin.tgz",
-										},
+										{Product: "bar", DistType: params.SLSDistType, TargetFile: "bar-sls.tgz"},
+										{Product: "foo", DistType: params.BinDistType, TargetFile: "foo-bin.tgz"},
+										{Product: "foo", DistType: params.SLSDistType, TargetFile: "foo-sls.tgz"},
 									},
 								},
 							},
--- a/apps/distgo/cmd/dist/dockertests/dist_docker_test.go
+++ b/apps/distgo/cmd/dist/dockertests/dist_docker_test.go
@@ -51,6 +51,12 @@ func main() {
 `
 	dockerfile = `FROM alpine:3.5
 `
+	slsDepDockerFile = `FROM alpine:3.5
+
+COPY foo-sls.tgz .
+COPY foo-bin.tgz .
+COPY bar-sls.tgz .
+`
 	dockerRepoPrefix = "test-docker-dist"
 )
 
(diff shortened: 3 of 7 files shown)


nmiyake:
Over-all approach looks good, but some questions/feedback.

---

Reviewed 7 of 7 files at r1.
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 43 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5PfT0TYvkfPDmo-Wj:-Kh5PfT0TYvkfPDmo-Wk:bajz8gr) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L43)):*
> ```Go
> 
> 	// link dependent artifacts into the context directory
> 	dockerDeps := dockerDistInfo.DistDeps
> ```

Unless you're using this variable multiple times, no need to declare separately -- you can just range over directly:

```go
for depProduct := range dockerDistInfo.DistDeps {
...
}
```

(currently you do use `dockerDeps` to do the access, but per the comment below makes more sense to do a `k, v` iteration style, in which case that access isn't needed anyway)

---

*[apps/distgo/cmd/dist/dockerdist.go, line 45 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5PGqIecV1UXxV9-AM:-Kh5PGqIecV1UXxV9-AN:bdrrxui) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L45)):*
> ```Go
> 	dockerDeps := dockerDistInfo.DistDeps
> 	for depProduct := range dockerDeps {
> 		for depDistType := range dockerDeps[depProduct] {
> ```

Since you're using the value of the key here (and multiple other times in the loop), makes more sense to use the loop iteration form that gives both the key and the value:

```go
for depProduct, distTypes := range dockerDeps {
...
```

Pretty sure this is true of the other iteration as well. Basically, if you ever iterate over a map and use both its key and its value, use the full form (if you're just using the key, then use this short form).

---

*[apps/distgo/cmd/dist/dockerdist.go, line 50 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5QQZaiO7LWoxHmERz:-Kh5QQZaiO7LWoxHmES-:bsenmlo) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L50)):*
> ```Go
> 				continue
> 			}
> 			depProductSpec := buildSpecWithDeps.DistDeps[depProduct]
> ```

Same as above -- unless you're accessing `depProductSpec` directly in the loop, just range directly over `buildSpecWithDeps.DistDeps[depProduct]` (no need to declare and assign separate variable)

---

*[apps/distgo/cmd/dist/dockerdist.go, line 52 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5aKeU3mBjXoqDAFl9:-Kh5aKeU3mBjXoqDAFlA:b-3dio19) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L52)):*
> ```Go
> 			depProductSpec := buildSpecWithDeps.DistDeps[depProduct]
> 			for _, depDist := range depProductSpec.Dist {
> 				if depDist.Info.Type() != depDistType {
> ```

What's this check about? Seems like a loop may not be the best approach here if only one iteration of it is going to match (unless I'm misreading something).

---

*[apps/distgo/cmd/dist/dockerdist.go, line 56 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5_vZc0hf1Ue3f-isu:-Kh5_vZc0hf1Ue3f-isv:b-b6uo5x) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L56)):*
> ```Go
> 				}
> 				artifactLocation := ArtifactPath(depProductSpec, depDist)
> 				outputLink := dockerDeps[depProduct][depDistType]
> ```

I think something like `destFileName` or `destPath` might make more sense.

Also, as currently written, looks like if the user specified a directory path (`foo/bar` or similar), this will fail if the directory path doesn't already exist. Probably fine to treat as user error (and this is probably true for other paths in config), but just flagging.

---

*[apps/distgo/cmd/dist/order_specs.go, line 40 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5Rgfbv2U7B__uOC9-:-Kh5RgfcWWoUQFoPrkWt:be4zbfe) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/order_specs.go#L40)):*
> ```Go
> 			for _, dep := range curDist.Info.Deps() {
> 				if dep == product {
> 					// allow self dependencies
> ```

Does it make sense to allow self dependencies? May be better to catch and treat as an error earlier on.

---

*[apps/distgo/params/dist.go, line 167 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5YD5ybCkMuPsIy7A_:-Kh5YD5ybCkMuPsIy7Aa:b-8e4v93) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/params/dist.go#L167)):*
> ```Go
> 	}
> 	d[product][distType] = outputArtifact
> 	return d
> ```

Returning the receiver makes sense for the builder pattern, but I don't think we really use this as a builder so don't think we really need to return the receiver here.

---

*[apps/distgo/params/dist.go, line 179 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5YesxFfaOQQUHNwZY:-Kh5YesxFfaOQQUHNwZZ:bdx89hj) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/params/dist.go#L179)):*
> ```Go
> 	ContextDir string
> 	// DistDeps is a map with key as a product and value as a map of dist types of that product
> 	// to the filename of the link that needs to be created in the context directory.
> ```

What does the `to the filename of the link that needs to be created in the context directory` part mean? Link to what?

Assume that you mean that the dist output for the dist type of the given product will be copied/linked to a file of the specified name within the context directory, but just wanted to verify.

Also, does this mean that you verified that each dist type can only produce a single file as output? If not, then how does this handle tasks that produce multiple files as output?

---

*[apps/distgo/params/product.go, line 73 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5_AC817oGekTITn-w:-Kh5_AC817oGekTITn-x:brpf7f8) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/params/product.go#L73)):*
> ```Go
> 		}
> 		for _, currDepProduct := range currDistCfg.Info.Deps() {
> 			currSpec, ok := allSpecs[currDepProduct]
> ```

These two for loops are effectively identical, so would make sense to decomp a function that unifies the common logic.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/97#-:-Kh5bUKIanJkt8gBwFml:b-755wv8)*

yvdinesh:
Review status: all files reviewed at latest revision, 9 unresolved discussions.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 43 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5PfT0TYvkfPDmo-Wj:-Kh8bD2exXVnWkL_cDAS:b-896fix) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L43)):*

Unless you're using this variable multiple times, no need to declare separately -- you can just range over directly:

```go
for depProduct := range dockerDistInfo.DistDeps {
...
}
```

(currently you do use `dockerDeps` to do the access, but per the comment below makes more sense to do a `k, v` iteration style, in which case that access isn't needed anyway)
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 45 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5PGqIecV1UXxV9-AM:-Kh8bdvmngZpizxSlifu:b-896fix) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L45)):*

Since you're using the value of the key here (and multiple other times in the loop), makes more sense to use the loop iteration form that gives both the key and the value:

```go
for depProduct, distTypes := range dockerDeps {
...
```

Pretty sure this is true of the other iteration as well. Basically, if you ever iterate over a map and use both its key and its value, use the full form (if you're just using the key, then use this short form).
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 50 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5QQZaiO7LWoxHmERz:-Kh8dKW0tw0WB8j2UXlC:b-q5adzc) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L50)):*

Same as above -- unless you're accessing `depProductSpec` directly in the loop, just range directly over `buildSpecWithDeps.DistDeps[depProduct]` (no need to declare and assign separate variable)
</blockquote></details>

I am accessing `depProductSpec` inside the loop.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 52 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5aKeU3mBjXoqDAFl9:-Kh8dbIhtIsA-F7cfhTK:bwq1a1x) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L52)):*

What's this check about? Seems like a loop may not be the best approach here if only one iteration of it is going to match (unless I'm misreading something).
</blockquote></details>

We iterate over all the dists and execute the function for only matching distType.
Ideally, we need a map from the the distType to slice of dist. But thought this is more easier to understand without incurring too much performance penality.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 56 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5_vZc0hf1Ue3f-isu:-Kh8d27NRqqD2IXgT02S:b-896fix) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/dockerdist.go#L56)):*

I think something like `destFileName` or `destPath` might make more sense.

Also, as currently written, looks like if the user specified a directory path (`foo/bar` or similar), this will fail if the directory path doesn't already exist. Probably fine to treat as user error (and this is probably true for other paths in config), but just flagging.
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/order_specs.go, line 40 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5Rgfbv2U7B__uOC9-:-Kh8cEfgBUHF3bo9DI25:b-iy69xa) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/cmd/dist/order_specs.go#L40)):*

Does it make sense to allow self dependencies? May be better to catch and treat as an error earlier on.
</blockquote></details>

self dependencies will be needed because i expect products to define an sls type and a docker type which depends on the sls type of the same product.

---

*[apps/distgo/params/dist.go, line 167 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5YD5ybCkMuPsIy7A_:-Kh8lNkDKCq8EN10GsT1:b-e1um7) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/params/dist.go#L167)):*

Returning the receiver makes sense for the builder pattern, but I don't think we really use this as a builder so don't think we really need to return the receiver here.
</blockquote></details>

changed the `DockerDistDep` to be of struct type to avoid this function.

---

*[apps/distgo/params/dist.go, line 179 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5YesxFfaOQQUHNwZY:-Kh8nADftSJrPnI_uFdi:b-k3bgx8) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/params/dist.go#L179)):*

What does the `to the filename of the link that needs to be created in the context directory` part mean? Link to what?

Assume that you mean that the dist output for the dist type of the given product will be copied/linked to a file of the specified name within the context directory, but just wanted to verify.

Also, does this mean that you verified that each dist type can only produce a single file as output? If not, then how does this handle tasks that produce multiple files as output?
</blockquote></details>

yes. thats the correct assumption. updated the comment to be more clear.
I verified that all the current dist types only output one artifact and it is determined by the function `ArtifactPath`

---

*[apps/distgo/params/product.go, line 73 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5_AC817oGekTITn-w:-Kh8sM0hjzxbU7bh47cJ:b-896fix) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/params/product.go#L73)):*

These two for loops are effectively identical, so would make sense to decomp a function that unifies the common logic.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/97)*

nmiyake:
Thanks, looks good -- a few more suggestions.

---

Reviewed 5 of 5 files at r2.

---

*[apps/distgo/params/product.go, line 73 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5_AC817oGekTITn-w:-Kh8v_eeyilzhMxp8YmC:b-ew3ms) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/params/product.go#L73)):*

Done.
</blockquote></details>

Can still be decomped further -- both `range` functions range through `[]string` inputs and assign into `buildDeps`, so it should be possible to factor out the entire `for` loop portion.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/97#-:-Kh8vypghZ1uLcbpqDZf:b-5x2sug)*

yvdinesh:

---

*[apps/distgo/params/product.go, line 73 at r1](https://reviewable.io:443/reviews/palantir/godel/97#-Kh5_AC817oGekTITn-w:-Kh93NY9XTLIm--De6Gb:b-s5dv8p) ([raw file](https://github.com/palantir/godel/blob/cb3fdb5e9c0b88bd686a1bae47cf9901cd301d53/apps/distgo/params/product.go#L73)):*

Can still be decomped further -- both `range` functions range through `[]string` inputs and assign into `buildDeps`, so it should be possible to factor out the entire `for` loop portion.
</blockquote></details>

one of assigns to `distDeps`. I have still refactored the for loop.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/97)*

nmiyake:
Reviewed 1 of 1 files at r3.

---

*[apps/distgo/params/product.go, line 147 at r3](https://reviewable.io:443/reviews/palantir/godel/97#-Kh9qXIz4-AbeA1UbKQe:-Kh9qXIz4-AbeA1UbKQf:baa6s9a) ([raw file](https://github.com/palantir/godel/blob/ba4962fc99560c4ec783a2de8dff96d3c8358854/apps/distgo/params/product.go#L147)):*
> ```Go
> }
> 
> func addDeps(deps map[string]ProductBuildSpec, depProudcts []string, spec ProductBuildSpec, allSpecs map[string]ProductBuildSpec) error {
> ```

Typo: should be `depProducts`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/97)*

nmiyake:
Minor typo, but should be good after that

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/97#-:-Kh9qggCsOh9rkfQN9zF:bnr3f4n)*

yvdinesh:

---

*[apps/distgo/params/product.go, line 147 at r3](https://reviewable.io:443/reviews/palantir/godel/97#-Kh9qXIz4-AbeA1UbKQe:-KhNrSC2IuticD3mz0QJ:b-896fix) ([raw file](https://github.com/palantir/godel/blob/ba4962fc99560c4ec783a2de8dff96d3c8358854/apps/distgo/params/product.go#L147)):*

Typo: should be `depProducts`
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/97)*

nmiyake:
:lgtm:

---

Reviewed 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/97#-:-KhOcXnx-eJXQxfyqqWr:bnfp4nl)*

PR #100, opened 2017-04-10
Improve output for building dist

yvdinesh:


nmiyake:
Some questions. Also, need to format files.

---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[apps/distgo/cmd/dist/bindist.go, line 26 at r1](https://reviewable.io:443/reviews/palantir/godel/100#-KhP8gsB95b4p-QLI8pf:-KhP8gsB95b4p-QLI8pg:b63uh3) ([raw file](https://github.com/palantir/godel/blob/9d8f715e8cd10ee4469932311e12eea809695a1e/apps/distgo/cmd/dist/bindist.go#L26)):*
> ```Go
> 	"github.com/pkg/errors"
> 
> 	"fmt"
> ```

This will fail -- imports are not organized

---

*[apps/distgo/cmd/dist/dist.go, line 82 at r1](https://reviewable.io:443/reviews/palantir/godel/100#-KhP96jnnYE2vgTgvvbf:-KhP96jnnYE2vgTgvvbg:bpi109u) ([raw file](https://github.com/palantir/godel/blob/4278af7acd4190893cc987461e58c89d364762cd/apps/distgo/cmd/dist/dist.go#L82)):*
> ```Go
> 		outputDir := path.Join(buildSpec.ProjectDir, currDistCfg.OutputDir)
> 
> 		fmt.Fprintf(stdout, "Creating distribution for %v at %v\n", buildSpec.ProductName, ArtifactPath(buildSpec, currDistCfg))
> ```

Why can't we keep this logic as-is? Doesn't the Docker dist also have a target output artifact where this logic would just work?

---

*[apps/distgo/cmd/dist/dockerdist.go, line 64 at r1](https://reviewable.io:443/reviews/palantir/godel/100#-KhP8vCQJww9AjB5nunT:-KhP8vCQJww9AjB5nunU:bz5p4hs) ([raw file](https://github.com/palantir/godel/blob/9d8f715e8cd10ee4469932311e12eea809695a1e/apps/distgo/cmd/dist/dockerdist.go#L64)):*
> ```Go
> 				}
> 				target := path.Join(contextDir, targetFile)
> 				if _, err := os.Stat(target); err == nil {
> ```

In the case that it does exist, is there any way to verify that it's valid/up-to-date?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100#-:-KhP9GiakMhZ8ZSIcXAK:bll87gy)*

yvdinesh:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[apps/distgo/cmd/dist/bindist.go, line 26 at r1](https://reviewable.io:443/reviews/palantir/godel/100#-KhP8gsB95b4p-QLI8pf:-KhPIiXQjim9DkeBw3xS:b-896fix) ([raw file](https://github.com/palantir/godel/blob/9d8f715e8cd10ee4469932311e12eea809695a1e/apps/distgo/cmd/dist/bindist.go#L26)):*

This will fail -- imports are not organized
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/dist.go, line 82 at r1](https://reviewable.io:443/reviews/palantir/godel/100#-KhP96jnnYE2vgTgvvbf:-KhPKdKM46ttEVSz4q4P:b-99n1j2) ([raw file](https://github.com/palantir/godel/blob/4278af7acd4190893cc987461e58c89d364762cd/apps/distgo/cmd/dist/dist.go#L82)):*

Why can't we keep this logic as-is? Doesn't the Docker dist also have a target output artifact where this logic would just work?
</blockquote></details>

I did not want to add docker to artifact path as it does not actually have a path and it might be confusing.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 64 at r1](https://reviewable.io:443/reviews/palantir/godel/100#-KhP8vCQJww9AjB5nunT:-KhPKqFEJX2Xb1xwhz4u:b-ca2at5) ([raw file](https://github.com/palantir/godel/blob/9d8f715e8cd10ee4469932311e12eea809695a1e/apps/distgo/cmd/dist/dockerdist.go#L64)):*

In the case that it does exist, is there any way to verify that it's valid/up-to-date?
</blockquote></details>

i cannot think of a way as there is no easy way to get this info from the docker image.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100)*

nmiyake:
Feedback on making the output logic more generic while handling the case of dist types that don't have an output artifact/path.

---

Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[apps/distgo/cmd/dist/dist.go, line 82 at r1](https://reviewable.io:443/reviews/palantir/godel/100#-KhP96jnnYE2vgTgvvbf:-KhQBbca9EHtkE2es7I0:b9zgmwx) ([raw file](https://github.com/palantir/godel/blob/4278af7acd4190893cc987461e58c89d364762cd/apps/distgo/cmd/dist/dist.go#L82)):*

I did not want to add docker to artifact path as it does not actually have a path and it might be confusing.
</blockquote></details>

Ah right. In that case, let's make this more generic. I think the formalized/generic version of this is something like the following:

* The artifact output path for a dist may be empt
* If it is empty, then the dist message will not include the destination

That way, this logic stays generic and the messages are consistent, and it will also apply equally to other future dist types that may not have artifact output paths.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 64 at r1](https://reviewable.io:443/reviews/palantir/godel/100#-KhP8vCQJww9AjB5nunT:-KhQBz4OdaPKFKyFOFte:bs4ixp3) ([raw file](https://github.com/palantir/godel/blob/9d8f715e8cd10ee4469932311e12eea809695a1e/apps/distgo/cmd/dist/dockerdist.go#L64)):*

i cannot think of a way as there is no easy way to get this info from the docker image.
</blockquote></details>

OK, fair enough.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100#-:-KhQC1S7-qvOXVN_5ijV:bjsnqg0)*

yvdinesh:
@nmiyake ready for review

nmiyake:
Reviewed 1 of 4 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[apps/distgo/cmd/dist/dist.go, line 83 at r3](https://reviewable.io:443/reviews/palantir/godel/100#-KhSu6JLRXAk4ZcRAz__:-KhSu6JLRXAk4ZcRAz_a:b7wcfub) ([raw file](https://github.com/palantir/godel/blob/a2c8943d929df3986fa7e136ced2dc6f5fbcfb3c/apps/distgo/cmd/dist/dist.go#L83)):*
> ```Go
> 
> 		if ArtifactPath(buildSpec, currDistCfg) != "" {
> 			fmt.Fprintf(stdout, "Creating distribution for %v at %v\n", buildSpec.ProductName, ArtifactPath(buildSpec, currDistCfg))
> ```

Still more duplication than I'd like. Let's unify:

```go
msg := fmt.Sprintf("Creating distribution for %s", buildSpec.ProductName")`
if distArtifactPath := ArtifactPath(buildSpec, currDistCfg); distArtifactPath != "" {
        msg += fmt.Sprintf(" at %s", distArtifactPath)
}
fmt.Fprintln(stdout, msg)
```

---

*[apps/distgo/cmd/dist/dist.go, line 230 at r3](https://reviewable.io:443/reviews/palantir/godel/100#-KhSv74J4SCKxA0iTfk4:-KhSv74J4SCKxA0iTfk5:b-8zarvj) ([raw file](https://github.com/palantir/godel/blob/a2c8943d929df3986fa7e136ced2dc6f5fbcfb3c/apps/distgo/cmd/dist/dist.go#L230)):*
> ```Go
> 		fileName = fmt.Sprintf("%v-%v-%v.x86_64.rpm", buildSpec.ProductName, buildSpec.ProductVersion, release)
> 	case params.DockerDistType:
> 		return ""
> ```

Let's add comment to clarify why this is the case (`docker build` does not produce a file as output)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100)*

nmiyake:
This is close, but a few more changes requested.

---

Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100#-:-KhSvFN7Z6ZDTNyEU4_z:bjuznn3)*

nmiyake:
:lgtm:

---

Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100#-:-KhSzrKKBa0BfO-fcnVs:bnfp4nl)*

yvdinesh:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[apps/distgo/cmd/dist/dist.go, line 82 at r1](https://reviewable.io:443/reviews/palantir/godel/100#-KhP96jnnYE2vgTgvvbf:-KhT0iYOe6wYi74cZgyq:b-896fix) ([raw file](https://github.com/palantir/godel/blob/4278af7acd4190893cc987461e58c89d364762cd/apps/distgo/cmd/dist/dist.go#L82)):*

Ah right. In that case, let's make this more generic. I think the formalized/generic version of this is something like the following:

* The artifact output path for a dist may be empt
* If it is empty, then the dist message will not include the destination

That way, this logic stays generic and the messages are consistent, and it will also apply equally to other future dist types that may not have artifact output paths.
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/dist.go, line 83 at r3](https://reviewable.io:443/reviews/palantir/godel/100#-KhSu6JLRXAk4ZcRAz__:-KhT0pqCfTbixDmM7U0l:b-896fix) ([raw file](https://github.com/palantir/godel/blob/a2c8943d929df3986fa7e136ced2dc6f5fbcfb3c/apps/distgo/cmd/dist/dist.go#L83)):*

Still more duplication than I'd like. Let's unify:

```go
msg := fmt.Sprintf("Creating distribution for %s", buildSpec.ProductName")`
if distArtifactPath := ArtifactPath(buildSpec, currDistCfg); distArtifactPath != "" {
        msg += fmt.Sprintf(" at %s", distArtifactPath)
}
fmt.Fprintln(stdout, msg)
```
</blockquote></details>

Done.

---

*[apps/distgo/cmd/dist/dist.go, line 230 at r3](https://reviewable.io:443/reviews/palantir/godel/100#-KhSv74J4SCKxA0iTfk4:-KhT0kwU_Af7SW8nXxhL:b-896fix) ([raw file](https://github.com/palantir/godel/blob/a2c8943d929df3986fa7e136ced2dc6f5fbcfb3c/apps/distgo/cmd/dist/dist.go#L230)):*

Let's add comment to clarify why this is the case (`docker build` does not produce a file as output)
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100)*

nmiyake:
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 30 at r4](https://reviewable.io:443/reviews/palantir/godel/100#-KhT0V0Q7f64ONyGXluW:-KhT0V0Q7f64ONyGXluX:b-mwtgfj) ([raw file](https://github.com/palantir/godel/blob/175b0454cf56906af28de6d40c2f1dbb32a2185a/apps/distgo/cmd/dist/dockerdist.go#L30)):*
> ```Go
> 
> func dockerDist(buildSpecWithDeps params.ProductBuildSpecWithDeps, distCfg params.Dist, stdout io.Writer) (Packager, error) {
> 	fmt.Fprintf(stdout, "Creating docker distribution for %v\n", buildSpecWithDeps.Spec.ProductName)
> ```

Now that we made the logic generic, doesn't this stutter?

Based on the code, I would think we would have:

```
Creating distribution for [product]
Creating docker distribution for [product]
```

both printed when we create a Docker distribution?

If you want to change the over-all `Creating distribution` output to include the dist type (so it would say "Creating docker distribution" or "Creating sls distribution" etc.) I'd be fine with that as well.

Whatever you ultimately decide, please try running locally to make sure that the output is sane.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 94 at r4](https://reviewable.io:443/reviews/palantir/godel/100#-KhT15fvGt69mJl5Scgg:-KhT15fwlIsQHLi_Bhbx:bdoxwmb) ([raw file](https://github.com/palantir/godel/blob/175b0454cf56906af28de6d40c2f1dbb32a2185a/apps/distgo/cmd/dist/dockerdist.go#L94)):*
> ```Go
> 	args = append(args, "--tag", tag)
> 	args = append(args, contextDir)
> 	fmt.Fprintf(stdout, "Building docker image %v\n", tag)
> ```

Not sure if this is really necessary -- seems more like debugging output.

Alternatively, if, in this case, the "tag" is acting as the de facto "destination" (where that's the output that people would want to see when building a Docker distribution, much in the same way that they would want to see the destination of a file), then I think we can try to design something to make the current unified message output this instead (which might make more sense).

Actually, I think that's the case -- the original intent of the output "Creating distribution for X at Y" was because people generally want to know where the output of a distribution will go. In this case, although it's not a file, the conceptual "output location" is the tag, and that's what people want to know.

I don't think it makes sense to make the tag the artifact location, since that is used by other things as well.

Do later docker tasks (dist or publish) need this tag? Trying to think of the best design/abstraction for this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100)*

yvdinesh:
apologize for not checking the actual output. Verified that output looks sane.

---

Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 30 at r4](https://reviewable.io:443/reviews/palantir/godel/100#-KhT0V0Q7f64ONyGXluW:-KhTBD_kUxgG4yI8kdkJ:b-ytshfk) ([raw file](https://github.com/palantir/godel/blob/175b0454cf56906af28de6d40c2f1dbb32a2185a/apps/distgo/cmd/dist/dockerdist.go#L30)):*

Now that we made the logic generic, doesn't this stutter?

Based on the code, I would think we would have:

```
Creating distribution for [product]
Creating docker distribution for [product]
```

both printed when we create a Docker distribution?

If you want to change the over-all `Creating distribution` output to include the dist type (so it would say "Creating docker distribution" or "Creating sls distribution" etc.) I'd be fine with that as well.

Whatever you ultimately decide, please try running locally to make sure that the output is sane.
</blockquote></details>

added the dist type when printing `Creating distribution` output in `dist.go`. removing this message.

---

*[apps/distgo/cmd/dist/dockerdist.go, line 94 at r4](https://reviewable.io:443/reviews/palantir/godel/100#-KhT15fvGt69mJl5Scgg:-KhTBUYUrZyGzqwSo8N5:b-lmlstj) ([raw file](https://github.com/palantir/godel/blob/175b0454cf56906af28de6d40c2f1dbb32a2185a/apps/distgo/cmd/dist/dockerdist.go#L94)):*

Not sure if this is really necessary -- seems more like debugging output.

Alternatively, if, in this case, the "tag" is acting as the de facto "destination" (where that's the output that people would want to see when building a Docker distribution, much in the same way that they would want to see the destination of a file), then I think we can try to design something to make the current unified message output this instead (which might make more sense).

Actually, I think that's the case -- the original intent of the output "Creating distribution for X at Y" was because people generally want to know where the output of a distribution will go. In this case, although it's not a file, the conceptual "output location" is the tag, and that's what people want to know.

I don't think it makes sense to make the tag the artifact location, since that is used by other things as well.

Do later docker tasks (dist or publish) need this tag? Trying to think of the best design/abstraction for this.
</blockquote></details>

ya, thats a good point. made `ArtifactPath` return the docker image tag.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100#-:-KhTAvt1H_ovk1SIIw70:bk81op4)*

nmiyake:
Discussed extensively in person, and at a higher level we'll likely refactor all the Docker stuff into another subcommand of `distgo`, but will merge this PR so we have it in history.

Can you revert to the r4 version of this? Basically, I don't want the Docker dist type to report an artifact path since that's conceptually incorrect. The output that's printed for Docker can just omit the destination/tag.

---

Reviewed 1 of 2 files at r5.
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100#-:-KhTP7MW4h9In1_wEcb1:b7le3tr)*

yvdinesh:
reverted.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100#-:-KhTb9oAA-JPCug3Hy5I:b-81taoq)*

nmiyake:
:lgtm:

---

Reviewed 2 of 2 files at r6.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/palantir/godel/100#-:-KhTbbIVa76jhFc_npCd:bnfp4nl)*

PR #124, opened 2017-04-24
automatically detect github upload URI

jdhenke:
this fails the `novendor` check on `github.com/google/go-querystring` and I don't know why...

### DIFF
--- a/apps/distgo/cmd/publish/github_publish.go
+++ b/apps/distgo/cmd/publish/github_publish.go
@@ -23,11 +23,9 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"reflect"
-	"strings"
 
 	"github.com/google/go-github/github"
-	"github.com/google/go-querystring/query"
+	"github.com/jtacoma/uritemplates"
 	"github.com/pkg/errors"
 	"golang.org/x/oauth2"
 	"gopkg.in/cheggaaa/pb.v1"
@@ -36,10 +34,9 @@ import (
 )
 
 type GitHubConnectionInfo struct {
-	APIURL    string
-	UploadURL string
-	User      string
-	Token     string
+	APIURL string
+	User   string
+	Token  string
 
 	Owner      string
 	Repository string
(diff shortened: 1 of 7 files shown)


nmiyake:
Pretty sure it's because this change removes the use of the "github.com/google/go-querystring/query": https://github.com/palantir/godel/pull/124/files#diff-d145779d1d8536b6f63adc070d379ef8L30

apps/distgo/cmd/publish/github_publish.go
```diff
@@ -102,13 +87,31 @@ func (g *GitHubConnectionInfo) Publish(buildSpec params.ProductBuildSpec, paths
...
 func (g *GitHubConnectionInfo) uploadFileAtPath(client *github.Client, release *github.RepositoryRelease, filePath string, stdout io.Writer) (string, error) {
+	// create asset upload URI using upload template provided in release creation
+	// response
+	//
+	// API: https://developer.github.com/v3/repos/releases/#response
+	//
+	// Examples:
+	//   - https://uploads.github.com/repos/octocat/Hello-World/releases/1/assets{?name,label}
+	//   - https://github.enterprise.host/api/uploads/repos/octocat/Hello-World/releases/1/assets{?name,label}
+	t, err := uritemplates.Parse(release.GetUploadURL())
```
nmiyake: Let's factor this out into a separate functions and then add tests with the example inputs above

apps/distgo/cmd/publish/github_publish.go
```diff
@@ -23,11 +23,9 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"reflect"
-	"strings"
 
 	"github.com/google/go-github/github"
-	"github.com/google/go-querystring/query"
```
nmiyake: OK tried locally, and looks like removing this from the vendor directory causes `extimport` to flag it because it is a real error (it's transitively used).

For now, we'll have to add to the `novendor` whitelist in `checks.yml`. This is issue https://github.com/palantir/checks/issues/16.

nmiyake:
I amended the commit to make the changes that I wanted based on first PR. @jdhenke take a look and let me know if this looks good -- if so, will go ahead and merge. Thanks for contributing the fix!

jdhenke:
@nmiyake LGTM - thanks!

PR #127, opened 2017-04-24
Only use base repo for remote tags in docker publish

yvdinesh:


yvdinesh:
actually, i dont think we need this fix, as we can also use base repo while building. closing this out.

PR #149, opened 2017-06-15
Add ability to specify SLS info to docker image

yvdinesh:


apps/distgo/config/config.go
```diff
@@ -256,11 +256,23 @@ type DockerDep struct {
 	TargetFile string `yaml:"target-file" json:"target-file"`
 }
 
+type DockerImageInfo struct {
```
nmiyake: Would move under declaration of `DockerImage` (and next to `SLSImageInfo`) for consistency

apps/distgo/config/config.go
```diff
@@ -256,11 +256,23 @@ type DockerDep struct {
...
-	ContextDir string      `yaml:"context-dir" json:"context-dir"`
-	Deps       []DockerDep `yaml:"dependencies" json:"dependencies"`
+	Repository string          `yaml:"repository" json:"repository"`
+	Tag        string          `yaml:"tag" json:"tag"`
+	ContextDir string          `yaml:"context-dir" json:"context-dir"`
+	Deps       []DockerDep     `yaml:"dependencies" json:"dependencies"`
+	Info       DockerImageInfo `yaml:"info" json:"info"`
+}
+
+type SLSImageInfo struct {
```
nmiyake: `SLSDockerImageInfo` would probably be a better name (wouldn't feel as strongly about this if this were in a `Docker` package or something, but since it's in a general `config` package I think it makes more sense to be explicit about the naming)

apps/distgo/params/docker.go
```diff
@@ -15,37 +15,44 @@
...
+	SetRepository(repo string)
+	SetDefaults(repo string, tag string)
+	Coordinates() (string, string)
+	ContextDirectory() string
+	Dependencies() []DockerDep
+	Build(buildSpec ProductBuildSpecWithDeps) error
 }
 
-type DockerImage struct {
+type SimpleDockerImage struct {
```
nmiyake: Not sure if `Simple` is the best naming here -- maybe `DefaultDockerImage` or `StandardDockerImage` instead? Not totally set on this, though (may be okay to leave as-is)

apps/distgo/params/manifest.go
```diff
@@ -0,0 +1,63 @@
...
+		missingRequired = append(missingRequired, "group-id")
+	}
+	if name == "" {
+		missingRequired = append(missingRequired, "product-name")
+	}
+	if version == "" {
+		missingRequired = append(missingRequired, "product-version")
+	}
+	if len(missingRequired) > 0 {
+		return "", errors.Errorf("required properties were missing: " + strings.Join(missingRequired, ", "))
```
nmiyake: Although string concat does work, since you're using `Errorf` more canonical to use format string:

`"required properties were missing: %s", strings.Join(...)"`

apps/distgo/config/config.go
yvdinesh: Done.

apps/distgo/config/config.go
yvdinesh: Renamed to `SLSDockerImageInfo`

apps/distgo/params/docker.go
yvdinesh: renamed to `DefaultDockerImage`

apps/distgo/params/manifest.go
yvdinesh: done.

yvdinesh:
@nmiyake ready for review.

apps/distgo/params/docker.go
```diff
@@ -15,37 +15,45 @@
...
-	for _, dep := range d {
-		if m[dep.Product] == nil {
-			m[dep.Product] = make(map[DockerDepType]string)
-		}
-		m[dep.Product][dep.Type] = dep.TargetFile
-	}
-	return m
+type DockerImage interface {
+	SetRepository(repo string)
+	SetDefaults(repo string, tag string)
```
nmiyake: For consecutive params of same type, you can declare type once:

```go
SetDefaults(repo, tag string)
```

apps/distgo/params/docker.go
```diff
@@ -60,7 +68,82 @@ type DockerImage struct {
...
+	}
+	args = append(args, "--label", fmt.Sprintf("%s=%s", ManifestLabel, base64.StdEncoding.EncodeToString([]byte(manifest))))
+	if _, err := os.Stat(configFile); err == nil {
+		content, err := ioutil.ReadFile(configFile)
+		if err != nil {
+			return errors.Wrapf(err, "Failed to read the file %s", configFile)
+		}
+		args = append(args, "--label", fmt.Sprintf("%s=%s", ConfigurationLabel, base64.StdEncoding.EncodeToString(content)))
+	} else {
+		fmt.Printf("Unable to find the configuration.yml file at: %v\n", configFile)
```
nmiyake: Shouldn't be using `Printf` here -- use `Fprintf` to a `stdout` stream instead.

Also, is this an error case where you should return an error? I would think that either it's an error (in which case you should return an error) or OK (in which case you can just proceed without printing any output)

apps/distgo/config/config.go
```diff
@@ -552,25 +564,68 @@ func (cfg *RPMDist) ToParams() params.RPMDistInfo {
...
+		if err := mapstructure.Decode(cfg.Info.Data, &slsInfo); err != nil {
+			return nil, errors.Wrap(err, "Failed to unmarshal the sls image info")
+		}
+		return &params.SLSDockerImage{
+			DefaultDockerImage: dockerImage,
+			ProuductType:       slsInfo.ProuductType,
+			GroupID:            slsInfo.GroupID,
+			Extensions:         slsInfo.Extensions,
+		}, nil
+	case "":
```
yvdinesh: Currently, the default docker image does not need any extra info. So, chose to return the default docker image in case the type is empty.

apps/distgo/params/docker.go
yvdinesh: done.

apps/distgo/params/docker.go
yvdinesh: Added this for debugging. Forgot to remove. Removing it.

yvdinesh:
@nmiyake ready for review!

apps/distgo/config/config.go
nmiyake: Ah, I see, and this is also in `ToParams`. OK, in that case feel better about this.

apps/distgo/params/docker.go
```diff
@@ -15,37 +15,45 @@
...
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path"
+
 	"github.com/pkg/errors"
 )
 
 type DockerDepType string
+type DockerImageType string
```
nmiyake: This type is defined but I don't ever see it used

apps/distgo/params/docker.go
```diff
@@ -15,37 +15,45 @@
...
 const (
-	DockerDepSLS    DockerDepType = "sls"
-	DockerDepBin    DockerDepType = "bin"
-	DockerDepRPM    DockerDepType = "rpm"
-	DockerDepDocker DockerDepType = "docker"
+	DockerDepSLS          DockerDepType = "sls"
+	DockerDepBin          DockerDepType = "bin"
+	DockerDepRPM          DockerDepType = "rpm"
+	DockerDepDocker       DockerDepType = "docker"
+	ManifestLabel         string        = "com.palantir.sls.manifest"
```
nmiyake: This variable and the two after it shouldn't have `string` type declared explicitly. Also add a newline between these and the `DockerDepType` block before it to make it clear that they're conceptually separate.

yvdinesh: done.

apps/distgo/params/docker.go
yvdinesh: removing the unused type.

yvdinesh:
@nmiyake not sure why the PR is updating from the base branch on the fork. I can see my commits on my fork.

PR #150, opened 2017-06-16
Repackage a newer version of errcheck in order to use file-backed exclusion

joshdk:
Specifically, to use the functionality provided by [kisielk/errcheck#116](https://github.com/kisielk/errcheck/pull/116).

### DIFF
--- a/apps/okgo/generated_src/internal/github.com/kisielk/errcheck/internal/errcheck/errcheck.go
+++ b/apps/okgo/generated_src/internal/github.com/kisielk/errcheck/internal/errcheck/errcheck.go
@@ -34,8 +34,9 @@ var (
 
 // UncheckedError indicates the position of an unchecked error return.
 type UncheckedError struct {
-	Pos	token.Position
-	Line	string
+	Pos		token.Position
+	Line		string
+	FuncName	string
 }
 
 // UncheckedErrors is returned from the CheckPackage function if the package contains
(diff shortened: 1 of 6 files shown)


palantirtech:
Thanks for your interest in [palantir/godel](https://github.com/palantir/godel), @joshdk! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

nmiyake:
Thanks for the PR!

I'm definitely not opposed to updating the vendor version here, but curious as to how you plan on using the `exclude` functionality and would be curious to see if there are any arguments we can provide (or existing functionality we can use) to encode the information as an option on errcheck in `exclude.yml`.

joshdk:
@nmiyake As an example of how I'm currently using this, I have the following contents in my `godel/config/check.yml`:

```yml
checks:
  errcheck:
    args:
      - "-ignore"
      - "github.com/mitchellh/colorstring:(Fprint|Print)(f|ln)?,bytes:Write(Byte|String)?"
```
The `-ignore` flag only accepts a single, comma delimited, regex for specifying functions to disregard. This regex can quickly get out of hand as the breadth and depth of ignored functions grows.

Instead, it would sometimes be preferable to use the more explicit (but more verbose) alternative, by putting the following in a text file and referencing it with the `-exclude` flag:

```
(*bytes.Buffer).Write
(*bytes.Buffer).WriteByte
(*bytes.Buffer).WriteString
github.com/mitchellh/colorstring.Fprint
github.com/mitchellh/colorstring.Fprintf
github.com/mitchellh/colorstring.Fprintln
github.com/mitchellh/colorstring.Print
github.com/mitchellh/colorstring.Printf
github.com/mitchellh/colorstring.Println
```

At a glance I can clearly see what is going on, and don't need to grok a regex.

PR #155, opened 2017-06-27
Add build args script for docker config

### LINKED ISSUE #154: distgo: pass arguments to docker build
I recently started using the `docker` task to build images and discovered it would be helpful to pass arguments to the `docker build` command. In my case, I wanted to use the `--build-arg` flag to pass in the version of the project being built.

The most flexible solution would probably be something similar to the `build-args-script` option in the `build` configuration, but I'd also be happy with a limited option to provide only the project version. Maybe a `version-arg: <argument name>` option or something?

yvdinesh:
Fixes #154


yvdinesh:
@nmiyake This PR adds the support for build args script for the docker task as well.

apps/distgo/cmd/build/build.go
```diff
@@ -245,21 +244,12 @@ func doBuildAction(action buildAction, buildSpec params.ProductBuildSpec, output
...
-	}
-
-	buildArgsString := strings.TrimSpace(stdoutBuf.String())
-	if buildArgsString != "" {
-		buildArgs := strings.Split(buildArgsString, "\n")
-		args = append(args, buildArgs...)
+	// get build args
+	buildArgs, err := script.GetBuildArgs(buildSpec, buildSpec.Build.BuildArgsScript)
+	if err != nil {
+		return errors.Wrapf(err, "")
```
nmiyake: If using `Wrapf`, the error should be annotated with extra context. If not, then the `error` can be returned directly without augmentation (this is a big difference between `pkg/errors` and `stacktrace`).

apps/distgo/params/docker/default_image.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (di *DefaultDockerImage) Build(buildSpec params.ProductBuildSpecWithDeps) error {
+	contextDir := path.Join(buildSpec.Spec.ProjectDir, di.ContextDir)
+	var args []string
+	args = append(args, "build")
+	args = append(args, "--tag", fmt.Sprintf("%s:%s", di.Repository, di.Tag))
+	buildArgs, err := script.GetBuildArgs(buildSpec.Spec, di.BuildArgsScript)
+	if err != nil {
+		return errors.Wrap(err, "")
```
nmiyake: Same

apps/distgo/params/docker/default_image.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	if len(buildArgs) > 0 {
```
nmiyake: No need for this check -- `buildArgs...` works properly on 0/`nil` slices

apps/distgo/pkg/script/script.go
```diff
@@ -92,3 +94,20 @@ func WriteAndExecute(buildSpec params.ProductBuildSpec, script string, stdOut, s
 	}
 	return nil
 }
+
+func GetBuildArgs(buildSpec params.ProductBuildSpec, script string) ([]string, error) {
+	stdoutBuf := bytes.Buffer{}
+	stderrBuf := bytes.Buffer{}
+	combinedBuf := bytes.Buffer{}
+	stdoutMW := io.MultiWriter(&stdoutBuf, &combinedBuf)
+	stderrMW := io.MultiWriter(&stderrBuf, &combinedBuf)
```
nmiyake: It seems like we only actually use the output of the `stdoutBuf` -- if that's the case, what's the purpose of having the combined buffer or stderr buffer?

yvdinesh: The `combinedBuf` is used in the returned error. There is a check on the `stderrBuf` to check if its empty. Otherwise, error is returned. This code is already existing code. I just moved it to a separate file.

apps/distgo/cmd/docker/build_test.go
```diff
@@ -35,6 +35,7 @@ import (
 	"github.com/palantir/godel/apps/distgo/cmd/dist"
 	"github.com/palantir/godel/apps/distgo/cmd/docker"
 	"github.com/palantir/godel/apps/distgo/params"
+	dockerparams "github.com/palantir/godel/apps/distgo/params/docker"
```
nmiyake: Why is this file moved to a subdirectory? I'm always a bit skeptical about introducing a package that is immediately aliased

apps/distgo/pkg/script/script.go
```diff
@@ -92,3 +94,20 @@ func WriteAndExecute(buildSpec params.ProductBuildSpec, script string, stdOut, s
...
+	if err := WriteAndExecute(buildSpec, script, stdoutMW, stderrMW, nil); err != nil || stderrBuf.String() != "" {
+		return nil, errors.Wrapf(err, "failed to execute build args script for %v: %v", buildSpec.ProductName, combinedBuf.String())
+	}
+
+	buildArgsString := strings.TrimSpace(stdoutBuf.String())
+	if buildArgsString != "" {
+		return strings.Split(buildArgsString, "\n"), nil
+	}
+	return []string{}, nil
```
nmiyake: This comment still stands -- should flip so that, if `buildArgsString == ""`, then we return `nil, nil`, and in the other case we do this operation.

apps/distgo/pkg/script/script.go
nmiyake: Ah I see yup.

apps/distgo/pkg/script/script.go
yvdinesh: done.

apps/distgo/cmd/docker/build_test.go
nmiyake: Yeah I don't think that's great rationale -- I think the underlying issue here points to an inversion of control.

By the design of the package, it looks like both the `config` and `params` packages are meant to store objects/types that are read from config and used by other tasks, but shouldn't really house logic itself. For example, if you look at the `dist` params, the logic associated with them does not live in the `params` package.

I think this code should be updated/refactored to move the logic out of the `params` package to address the underlying dependency graph issue.

yvdinesh: refactored the code. Now the params package does not have the logic for building the image. moved it to the docker package. This now consistent with dist and build packages.

yvdinesh:
@nmiyake ready for review.

PR #161, opened 2017-07-19
Update IDEA command to support Gogland

nmiyake:


bmoylan:
thanks! sorry I didn't get to this, but I'm happy to try it out.

gdearment:
Would love to try this out, as the latest release of Idea no longer supports go, so i was tricked into moving to gogland

nmiyake:
Same -- that's what pushed me to implement this :)

nmiyake:
Will give another test run locally and if everything checks out I'll tag a release so that projects can bump to this

PR #162, opened 2017-07-27
Make docker build skip dist task for up to date products

yvdinesh:


apps/distgo/cmd/dist/dist.go
```diff
@@ -30,11 +30,48 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	productsSlice := []string{}
+	productsMap := make(map[string]struct{})
+	for _, curSpecWithDeps := range buildSpecsWithDeps {
+		for _, curDistCfg := range curSpecWithDeps.Spec.Dist {
+			artifactPaths := FullArtifactsPaths(ToDister(curDistCfg.Info), curSpecWithDeps.Spec, curDistCfg)
+			for _, artifactPath := range artifactPaths {
+				if fi, err := os.Stat(artifactPath); err == nil {
```
nmiyake: Invert check here to prevent further indent:

```go
fi, err := os.Stat(artifactPath)
if err != nil {
   continue
}
```

apps/distgo/cmd/dist/dist.go
```diff
@@ -30,11 +30,48 @@ import (
...
+					if goFiles, err := imports.AllFiles(
+						path.Join(
+							curSpecWithDeps.Spec.ProjectDir,
+							curSpecWithDeps.Spec.Build.MainPkg,
+						),
+					); err == nil {
+						if newerThan, err := goFiles.NewerThan(fi); err == nil && !newerThan {
```
nmiyake: Same as above

apps/distgo/cmd/dist/dist.go
```diff
@@ -30,11 +30,48 @@ import (
...
+							// if the dist artifact for the product already exists and none of the source files for the
+							// product are newer than the dist artifact, consider spec up-to-date
+							continue
```
nmiyake: Macro question: is it not possible to delegate to `build.RequiresBuild` instead of duplicating the logic here?

apps/distgo/cmd/dist/dist.go
yvdinesh: we cannot directly continue. The error not being nil means that the artifact does not exist at the destination path and you need to build it. So, we need to add it to the productsMap.

apps/distgo/cmd/dist/dist.go
yvdinesh: done.

apps/distgo/cmd/dist/dist.go
```diff
@@ -35,6 +35,34 @@ import (
 	"github.com/palantir/godel/apps/distgo/pkg/slsspec"
 )
 
+func RequiresDist(products []string, cfg params.Project, wd string) ([]string, error) {
+	buildSpecsWithDeps, err := build.SpecsWithDepsForArgs(cfg, products, wd)
+	if err != nil {
+		return nil, err
+	}
+	productsSlice := []string{}
```
nmiyake: Use `var productsSlice []string` style for declaring empty slices

apps/distgo/cmd/dist/dist.go
```diff
@@ -35,6 +35,34 @@ import (
...
+			for _, artifactPath := range artifactPaths {
+				if _, err := os.Stat(artifactPath); err == nil {
+					buildInfo := build.RequiresBuild(curSpecWithDeps, nil)
+					if len(buildInfo.Specs()) == 0 {
+						continue
+					}
+				}
+				productsMap[curSpecWithDeps.Spec.ProductName] = struct{}{}
+			}
+
```
nmiyake: Remove newline

nmiyake:
Changes were minor so I amended the PR. Thanks!

PR #168, opened 2017-08-09
Make gunit test deal with keys with capitalization correctly

### LINKED ISSUE #167: gunit: default test behavior does not work properly if a tag key contains capitalization
If a test key is capitalized, running tests without tags does not work properly.

Configuration like this works:

```
tags:
  integration:
    names:
      - "^integration$"
```

In this case, running `./godelw test` will run all test except those matching `^integration$`.

However, if the same configuration key contains capitalization:

```
tags:
  Integration:
    names:
      - "^integration$"
```

Then this no longer works -- running `./godelw test` runs all tests (it does not exclude the tests that match the tags). This is true for all tags if any key contains capitalization.

nmiyake:
Fixes #167


nmiyake:
@ryanmcnamara

Issue #237, opened 2017-10-19
Do not truncate startup log in default SLS init start


jonsyu1:
The default `slsDistInitSh` [truncates the startup log](https://github.com/palantir/godel/blob/0.28.0/apps/distgo/cmd/dist/slsdist.go#L93) on init start. If the application dumps a crashlog to stdout on exit, the next startup of the application will truncate the crashlog.

nmiyake:
This operation is now provided by an asset -- if we want to fix this issue, please open an issue on the repository for the asset.

PR #238, opened 2017-10-19
Do not truncate startup logs

jonsyu1:
Concatenate to the startup logs instead of truncating.

Fixes: #237

### DIFF
--- a/apps/distgo/cmd/dist/slsdist.go
+++ b/apps/distgo/cmd/dist/slsdist.go
@@ -90,7 +90,7 @@ start)
     mkdir -p "var/log"
     mkdir -p "var/run"
 
-    PID=$($SERVICE_CMD > var/log/$SERVICE-startup.log 2>&1 & echo $!)
+    PID=$($SERVICE_CMD >> var/log/$SERVICE-startup.log 2>&1 & echo $!)
     echo $PID > $PIDFILE
     sleep 1
     if is_process_active $PID; then


bmoylan:
I disagree with this -- it will create an infinitely growing file with nothing managing/rotating it.

jonsyu1:
The application should not concern itself with log rotation because this may vary from system to system. Log rotation is most commonly handled by [logrotate](https://linux.die.net/man/8/logrotate) and run as part of a daily cron job.

Without this change, we lose startup logs whenever we call start in the default init script. This problem is exacerbated by automatic start triggers (especially those with retries), hiding critical crashlogs like golang panic output. We can work around this by providing our own init script which concatenates instead of truncating, but this seems like a change worth making to the default implementation.

divath:
I agree with jyu.

1. What is the point of logging if we lose logs when they most useful - that is your program crashed and was restarted?

2. It is not possible to reliably log certain fatal crashes (e.g. panics) without runtime hacks redirecting all stderr to other files. Just today my program crashed (most likely some panic) and I don't know what was the cause because it was overridden.

3. If logs growing is an issue, then your service should provide custom logging with log rotation / collection, and most services do exactly that. Startup logs are basically a place where all unhandled errors go.

nmiyake:
Disters are now assets that have been factored out of the core code base. Closing this PR -- if this change is still desired, a PR should be opened on the asset project.

PR #245, opened 2017-11-20
Backing up startup log file

divath:
Keeping at least MAX_STARTUP_LOG_FILE_SIZE startup logs


divath:
@nmiyake @bmoylan @jonsyu1 

This is a middle-ground between current implementation and #238. 
It is a fairly naive log rotation in bash, but it should do the job in most cases.

bmoylan:
I'm more ok with this since it does not allow for unbounded growth

divath:
Any updates?

PR #256, opened 2017-12-03
distgo: Use build directory for temporary scripts

### LINKED ISSUE #255: distgo: scripts result in dirty working tree
When distgo creates the temporary bash scripts in [script.go](https://github.com/palantir/godel/blob/1.0.0/apps/distgo/pkg/script/script.go#L32) it uses the project directory for the tmpfile which is hard to gitignore. This means that during the script's execution, the working directory is dirty (which results in incorrect `project-version` commands in the script).

bmoylan:
Fixes #255

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/godel/256)

### DIFF
--- a/apps/distgo/pkg/script/script.go
+++ b/apps/distgo/pkg/script/script.go
@@ -29,7 +29,7 @@ import (
 )
 
 func Write(buildSpec params.ProductBuildSpec, script string) (name string, cleanup func() error, rErr error) {
-	tmpFile, err := ioutil.TempFile(buildSpec.ProjectDir, "")
+	tmpFile, err := ioutil.TempFile(buildSpec.Build.OutputDir, "distgo-script-")
 	if err != nil {
 		return "", nil, errors.Wrapf(err, "Failed to create script file")
 	}


apps/distgo/pkg/script/script.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 func Write(buildSpec params.ProductBuildSpec, script string) (name string, cleanup func() error, rErr error) {
-	tmpFile, err := ioutil.TempFile(buildSpec.ProjectDir, "")
+	tmpFile, err := ioutil.TempFile(buildSpec.Build.OutputDir, "distgo-script-")
```
bmoylan: Is this directory guaranteed to exist?

bmoylan: According to the test failures, apparently not. Is there a reason we don't want to use the host's TMDIR by leaving this argument empty?

nmiyake:
This has been fixed in gödel 2.0, so closing PR.

PR #416, opened 2018-07-31
Upgrade distgo to 1.13.0

gregakinman:
@nmiyake I took the checksums from https://bintray.com/palantir/releases/distgo which is I assume the correct place


nmiyake:
Thanks!

Issue #427, opened 2018-10-05
Support plugin overrides


nmiyake:
Currently, if a godel configuration uses a configuration provider that provides a plugin and a different version of the same plugin is specified in local configuration, they are reported as a conflict:

```
Error: 2 plugins had compatibility issues:
    com.palantir.<group>:<plugin>:1.1.0:
        different version of the same plugin
    com.palantir.<group>:<plugin>:2.0.0-rc2:
        different version of the same plugin
```

It would be nice to expose a mechanism that can be used to specify that the version specified locally should be used.

The proposed design is as follows:

* A single provider cannot specify multiple versions of the same plugin
* The local "plugins" block cannot specify multiple versions of the same plugin
* Plugins that are specified locally (in the `godel.yml`) can specify an "override: true" property
* If this property exists, then the resolver specified locally will be used (it will override any value provided by a config provider)

TBD:
* If multiple providers are specified and they define different versions of the same plugin and the local definition specifies the same plugin with an "override", should the override be used or should the conflict between providers be flagged?
  * Probably use the override -- this provides an explicit mechanism to resolve conflicts between providers if necessary
* Should a similar mechanism be introduced for default tasks?

PR #446, opened 2019-03-11
Allow plugin overrides

k-simons:
- Fix https://github.com/palantir/godel/issues/427

### DIFF
--- a/framework/godel/config/internal/v0/godel.go
+++ b/framework/godel/config/internal/v0/godel.go
@@ -76,7 +76,7 @@ type PluginsConfig struct {
 type SinglePluginConfig struct {
 	// LocatorWithResolverConfig stores the locator and the resolver for the plugin.
 	LocatorWithResolverConfig `yaml:",inline,omitempty"`
-	// LocatorWithResolverConfig stores the locator and the resolver for the plugin.
+	// Override is used to override a plugin found in the ConfigProvider that has the same locator as this plugin
 	Override bool `yaml:"override,omitempty"`
 	// Assets stores the locators and resolvers for the assets for this plugin.
 	Assets []LocatorWithResolverConfig `yaml:"assets,omitempty"`
--- a/framework/godel/config/internal/v0/godel.go
+++ b/framework/godel/config/internal/v0/godel.go
@@ -76,6 +76,8 @@ type PluginsConfig struct {
 type SinglePluginConfig struct {
 	// LocatorWithResolverConfig stores the locator and the resolver for the plugin.
 	LocatorWithResolverConfig `yaml:",inline,omitempty"`
+	// LocatorWithResolverConfig stores the locator and the resolver for the plugin.
+	Override bool `yaml:"override,omitempty"`
 	// Assets stores the locators and resolvers for the assets for this plugin.
 	Assets []LocatorWithResolverConfig `yaml:"assets,omitempty"`
 }
--- a/framework/godellauncher/param.go
+++ b/framework/godellauncher/param.go
@@ -38,7 +38,9 @@ type PluginsParam struct {
 
 type SinglePluginParam struct {
 	artifactresolver.LocatorWithResolverParam
-	Assets []artifactresolver.LocatorWithResolverParam
+	Assets           []artifactresolver.LocatorWithResolverParam
+	Override         bool
+	FromPluginConfig bool
 }
 
 // ConfigDirPath returns the path to the gödel configuration directory given the path to the project directory. Returns
--- a/framework/plugins/plugin.go
+++ b/framework/plugins/plugin.go
@@ -172,6 +172,7 @@ func groupByGroupAndProduct(plugins []godellauncher.SinglePluginParam) map[strin
 // dedupePlugins will remove any plugins that added by the configuration provider iff there is a plugin
 // present from the plugin portion of configuration that contains the same group and product name
 func dedupePlugins(plugins []godellauncher.SinglePluginParam) []godellauncher.SinglePluginParam {
+	fmt.Println("RUNNING DEDUPE")
 	var allPlugins []godellauncher.SinglePluginParam
 	pluginMap := groupByGroupAndProduct(plugins)
 	for _, pluginsForProductAndGroup := range pluginMap {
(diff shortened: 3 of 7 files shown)


nmiyake:
Thanks for the PR! Will review tonight or tomorrow.

framework/godel/config/internal/v0/godel.go
```diff
@@ -76,6 +76,8 @@ type PluginsConfig struct {
 type SinglePluginConfig struct {
 	// LocatorWithResolverConfig stores the locator and the resolver for the plugin.
 	LocatorWithResolverConfig `yaml:",inline,omitempty"`
+	// LocatorWithResolverConfig stores the locator and the resolver for the plugin.
```
nmiyake: Comment is copy-pasted -- should be updated to describe the `Override` parameter.

framework/plugins/plugin.go
```diff
@@ -159,6 +160,51 @@ func resolvePlugins(pluginsDir, assetsDir, downloadsDir string, osArch osarch.OS
 	return nil, errors.New(strings.Join(errStringsParts, "\n"+strings.Repeat(" ", pluginsinternal.IndentSpaces)))
 }
 
+func groupByProductAndGroup(plugins []godellauncher.SinglePluginParam) map[string][]godellauncher.SinglePluginParam {
+	pluginByGroupAndProduct := make(map[string][]godellauncher.SinglePluginParam)
+	for _, currPlugin := range plugins {
+		groupKey := currPlugin.LocatorWithChecksums.GroupAndProductString()
+		_, ok := pluginByGroupAndProduct[groupKey]
```
nmiyake: No need to do this check -- appending to `nil` is valid, so you can just do:

```go
groupKey := currPlugin.LocatorWithChecksums.GroupAndProductString()
pluginByGroupAndProduct[groupKey] = append(pluginByGroupAndProduct[groupKey], currPlugin)
```

framework/plugins/plugin.go
```diff
@@ -159,6 +160,51 @@ func resolvePlugins(pluginsDir, assetsDir, downloadsDir string, osArch osarch.OS
 	return nil, errors.New(strings.Join(errStringsParts, "\n"+strings.Repeat(" ", pluginsinternal.IndentSpaces)))
 }
 
+func groupByProductAndGroup(plugins []godellauncher.SinglePluginParam) map[string][]godellauncher.SinglePluginParam {
```
nmiyake: Would rename this to `groupByGroupAndProduct` just to keep this consistent (the other variables in this function go by `GroupAndProduct`)

framework/godellauncher/param.go
```diff
@@ -38,7 +38,9 @@ type PluginsParam struct {
 
 type SinglePluginParam struct {
 	artifactresolver.LocatorWithResolverParam
-	Assets []artifactresolver.LocatorWithResolverParam
+	Assets           []artifactresolver.LocatorWithResolverParam
+	Override         bool
+	FromPluginConfig bool
```
nmiyake: Looking at the code, I don't see a case in which this won't be true (it appears that every instance of constructing a `SinglePluginParam` comes from config). What's the intent behind the variable?

framework/plugins/plugin.go
```diff
@@ -159,6 +160,51 @@ func resolvePlugins(pluginsDir, assetsDir, downloadsDir string, osArch osarch.OS
...
+		if !ok {
+			pluginByGroupAndProduct[groupKey] = make([]godellauncher.SinglePluginParam, 0)
+		}
+		singlePluginList := pluginByGroupAndProduct[groupKey]
+		pluginByGroupAndProduct[groupKey] = append(singlePluginList, currPlugin)
+	}
+	return pluginByGroupAndProduct
+}
+
+// dedupePlugins will remove any plugins that added by the configuration provider iff there is a plugin
```
nmiyake: Typos in comment

framework/plugins/plugin.go
```diff
@@ -159,6 +160,51 @@ func resolvePlugins(pluginsDir, assetsDir, downloadsDir string, osArch osarch.OS
...
+// present from the pluging portion of configuration that contains the same group and product name
+func dedupePlugins(plugins []godellauncher.SinglePluginParam) []godellauncher.SinglePluginParam {
+	var allPlugins []godellauncher.SinglePluginParam
+	pluginMap := groupByProductAndGroup(plugins)
+	for _, pluginsForProductAndGroup := range pluginMap {
+		noOverride := noOverridePresent(pluginsForProductAndGroup)
+		if noOverride {
```
nmiyake: The loop logic is the same over-all, so would push this check down into the loop logic instead:

```go
for _, singlePlugin := range pluginsForProductAndGroup {
	if noOverride || singlePlugin.FromPluginConfig {
}
```

framework/plugins/plugin.go
k-simons: Fixed "pluging". "iff" was intentional to be "if and only if"

framework/godellauncher/param.go
k-simons: Maybe I misunderstood how it was being used. I only set it to try in:

```
func (c *SinglePluginConfig) ToParam() (godellauncher.SinglePluginParam, error) {
```
The idea was it would be true if the `SinglePluginParam` came from a `SinglePluginConfig` but default to false in all other cases.

The reason that we need this is because (per the ticket) we want to error if there are 2 identical plugins found in the "plugins" portion of configuration. So we need to know if they came from the `plugin` configuration of the `configprovider`

nmiyake:
Thanks for adding the integration test! Took another look based on this, and I think that it will actually be easiest to perform the de-duping when the configuration from the config providers and local config is combined.

Specifically, at https://github.com/palantir/godel/blob/723a5b3f06a1384237c9386e44d31a3bae360e7b/main.go#L82, we have clear separation between `providedConfigs` (config from config providers) and `godelCfg.TasksConfig` (config from the local configuration file).

I think it should be pretty easy to do override resolution at this point. Specifically, we can:

1. Ensure that `providedConfigs` does not specify any "override" properties (if they do, we can just ignore them/set to false)
2. Do a partial resolution: for each plugin config with "override" specified, remove all plugin configurations with the same `group:name` combination from the provided plugin configuration

That should be pretty easy to do, and I think the rest of the dedupe logic that runs later will handle the other cases (like if the configuration specifies multiple overrides).

This should also simplify the code, as I don't think we need to track `override` at the param level or do any other deduplication later.

I'd be happy to handle the implementation of this. Thanks for the PR -- the initial work done here and the integration test that was written is very helpful.

nmiyake:
Closing now that #457 has merged.

Thanks a lot for the PR -- many parts of the eventual implementation was based on your PR, and the feature wouldn't have merged on this timeline without it!

PR #532, opened 2021-03-06
Properly close TGZ reader after opening it

nmiyake:
## Before this PR
TGZ file is opened for reading but not closed.

## After this PR

==COMMIT_MSG==
Fixes bug where artifact resolution logic would open a TGZ file to read but would not close the file.
==COMMIT_MSG==


tabboud (approved):
👍

PR #534, opened 2021-03-08
Use filelock when resolving artifacts

### LINKED ISSUE #533: Artifact resolution can fail when performed concurrently
## What happened?
When running two different copies of godel concurrently with an empty `GODEL_HOME`, they often fail with errors like the following:

```
Error: failed to resolve 2 plugin(s):
    failed to compute checksum for artifact at local-test-godel-home/downloads/com.palantir.godel-goland-plugin-goland-plugin-1.2.0.tgz: failed to read tar file entry: flate: corrupt input before offset 3489784
    failed to compute checksum for artifact at local-test-godel-home/downloads/com.palantir.okgo-check-plugin-1.9.0.tgz: failed to read tar entry: archive/tar: invalid tar header
```

Artifact resolution should succeed even if multiple copies of godel are preforming them at the same time

nmiyake:
## Before this PR
If multiple godel processes were resolving artifacts concurrently, the operations would often fail because one process would attempt to read a file that was being actively written by another process.

## After this PR

==COMMIT_MSG==
Fixes issue where artifact resolution could fail if multiple copies of godel were resolving artifacts concurrently. Note that, in order for this fix to work, all of the copies of godel that are running concurrently must have this fix.

Uses file locks to perform operations that involve downloading, checksumming
and extracting artifacts to ensure that running the operations from
multiple different copies of godel do not conflict.
==COMMIT_MSG==

Fixes #533


tabboud (approved):
Nice find!

PR #538, opened 2021-03-09
Remove individual file locks

nmiyake:
## After this PR

==COMMIT_MSG==
Removes individual file locks as they are no longer necessary now that godel-resolver.lock is used.
==COMMIT_MSG==


tabboud (approved):
👍

PR #540, opened 2021-03-15
Use consistent indent spacing in godelw

nmiyake:


tabboud (approved):
👍 

If only there was an automated way to enforce this as well. It doesn't look like shellcheck currently does.

PR #571, opened 2021-04-14
Add support for arm64 architectures

### LINKED ISSUE #570: Add support for arm64 architecture
## What happened?

When trying to use godel to build a multiarch package, it fails since the godel binary downloads amd64 version instead of arm64.
```
$ ./godelw 
Downloading https://github.com/palantir/godel/releases/download/v2.36.0/godel-2.36.0.tgz to /home/ruckc/.godel/downloads/godel-2.36.0-1990460.tgz...
/home/ruckc/.godel/downloads/godel-2.36.0-1990460. 100%[===============================================================================================================>]  11.68M  13.6MB/s    in 0.9s    
./godelw: line 147: /home/ruckc/.godel/tmp_7gRNr8/godel-2.36.0/bin/linux-amd64/godel: cannot execute binary file: Exec format error
Version reported by godel executable did not match expected version: expected "godel version 2.36.0", was ""
```

Godel to run

ruckc:
## Before this PR
An initial attempt at a solution to support arm64. Fixes #570.

## After this PR
Support building/uploading arm64 binaries and have godelw detect and fetch appropriate binaries.

## Possible downsides?
Unknown `arch` command output on darwin arm64 devices to test against... seeded `$ARCH` in `godelw` with `aarch64` like linux reports.


palantirtech:
Thanks for your interest in [palantir/godel](https://github.com/palantir/godel), @ruckc! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

godel/config/dist-plugin.yml
```diff
@@ -8,8 +8,12 @@ products:
       os-archs:
       - os: darwin
         arch: amd64
+      - os: darwin
+        arch: arm64
       - os: linux
         arch: amd64
+      - os: linux
```
nmiyake, line 15: This is probably the right thing to do long-term, but would like to scope this PR to just adding darwin-arm64 support

godel/config/dist-plugin.yml
```diff
@@ -52,6 +56,7 @@ products:
                 local os_arch=$1
                 local file=$2
                 local checksum=$(compute_sha256 "$DIST_WORK_DIR/bin/$os_arch/godel")
+                echo "Substituting checksum $os_arch $file $checksum"
```
nmiyake, line 59: Remove echo

resources/wrapper/godelw
```diff
@@ -161,23 +164,35 @@ GODEL_BASE_DIR=${GODEL_HOME:-$HOME/.godel}
 # determine OS
 OS=""
 EXPECTED_CHECKSUM=""
-case "$(uname)" in
-    Darwin*)
+case "$(uname)-$(arch)" in
```
nmiyake, line 167: Let's use `$(uname -m)` here -- `arch` outputs `i386` on Intel Macs, including 64-bit ones (see https://apple.stackexchange.com/questions/140651/why-does-arch-output-i386 and verified by testing locally as well), while `uname -m` outputs `x86_64` on 64-bit Intel macs and `arm64` on M1 Macs.

resources/wrapper/godelw
```diff
@@ -161,23 +164,35 @@ GODEL_BASE_DIR=${GODEL_HOME:-$HOME/.godel}
...
+    Darwin*-x86_64)
```
nmiyake, line 168: Based on testing, I think it's safe to remove the `*` and just have this be `Darwin-x86_64`

resources/wrapper/godelw
```diff
@@ -161,23 +164,35 @@ GODEL_BASE_DIR=${GODEL_HOME:-$HOME/.godel}
...
         OS=darwin
-        EXPECTED_CHECKSUM=$DARWIN_CHECKSUM
+        ARCH=amd64
+        EXPECTED_CHECKSUM=$DARWIN_AMD64_CHECKSUM
         ;;
-    Linux*)
+    Darwin*-aarch64)
```
nmiyake, line 173: Should be `Darwin-arm64`

nmiyake (changes requested):
Thanks for the PR, this looks great!

Some feedback: main points are request that we limit this PR to adding `darwin-arm64` and to use `uname -m` to determine architecture rather than `arch` to deal with the quirks of the macOS implementation

nmiyake:
The `integration-alpine-go-prev` check is failing because it runs integration tests in a Go 1.15 environment -- the integration test run `./godelw dist` which runs `./godelw build` which attempts to build the binary for darwin-arm64, but that support was only added in Go 1.16 so it fails.

This issue will resolve automatically once Go 1.17 becomes the current version, as the previous version will then become Go 1.16. However, I believe that Go 1.17 won't be released until August or September, which is still a ways out... This is a pretty infrequent issue (new OS/Arch that we want to target being introduced in a new version of Go), so I think I'd be open to a temporary workaround like disabling the integration tests for Go 1.15 and opening an issue to reenable once Go 1.17 is released (other things would include adding something in the integration test itself to change what is built, but think this would be hard to do cleanly).

ruckc:
I also need the linux-arm64 primarily.  I was adding darwin-arm64 only because you mentioned it (and it makes sense).  I'll try updating the PR tomorrow with the changes, especially `uname -m` which returns `aarch64` on linux-arm64.

nmiyake:
Ah OK, thanks for the context -- that's good to know, since I'll need to update publishing on all the other plugins/assets as well (but happy to do so if there's a use case, which it looks like there is :).

resources/wrapper/godelw
```diff
@@ -161,23 +164,35 @@ GODEL_BASE_DIR=${GODEL_HOME:-$HOME/.godel}
...
-    Darwin*)
+case "$(uname)-$(uname -m)" in
+    Darwin-x86_64)
         OS=darwin
-        EXPECTED_CHECKSUM=$DARWIN_CHECKSUM
+        ARCH=amd64
+        EXPECTED_CHECKSUM=$DARWIN_AMD64_CHECKSUM
         ;;
-    Linux*)
+    Darwin-aarch64)
```
nmiyake, line 173: This should be `Darwin-arm64` (verified locally that output of `uname -m` on an M1 mac is `arm64`)

resources/wrapper/godelw
```diff
@@ -161,23 +164,35 @@ GODEL_BASE_DIR=${GODEL_HOME:-$HOME/.godel}
...
+        OS=darwin
+        ARCH=arm64
+        EXPECTED_CHECKSUM=$DARWIN_ARM64_CHECKSUM
+        ;;
+    Linux-x86_64)
+        OS=linux
+        ARCH=amd64
+        EXPECTED_CHECKSUM=$LINUX_AMD64_CHECKSUM
+        ;;
+    Linux-aarch64)
```
nmiyake, line 183: Can you verify/confirm that output of `uname -m` on the linux arm environment you are running returns `aarch64`?

nmiyake (changes requested):

resources/wrapper/godelw
ruckc: Yes it does, namely my ubuntu 20.04 on raspberry pi 

```sh
ruckc@pi50:~$ uname
Linux
ruckc@pi50:~$ uname -m
aarch64
ruckc@pi50:~$ uname -a
Linux pi50 5.4.0-1032-raspi #35-Ubuntu SMP PREEMPT Fri Mar 19 20:52:40 UTC 2021 aarch64 aarch64 aarch64 GNU/Linux

```

resources/wrapper/godelw
ruckc: made this change

nmiyake (approved):
Thanks for the contribution!

nmiyake:
The [v2.37.0](https://github.com/palantir/godel/releases/tag/v2.37.0) contains this change along with the update to the default tasks. I tried it out on an M1 mac (darwin-arm64) and it appeared to work so I assume it should work on linux-arm64 as well, but let me know if you encounter any issues.

PR #593, opened 2021-06-18
Reverse order of computed resolvers list

akarnani:
## Before this PR

Previously the hard-coded default resolver was used first.  This meant that even if resolvers were specified in config, godel would still attempt to reach the hard coded default (github.com).  In environments without internet access this behavior is undesirable.

## After this PR

Now the configured resolvers are checked first and the default is used as a fallback.

## Possible downsides?

If the configured resolvers do not have most of the plugins/assets this change will cause more resolver misses.  However, these misses will only make the initial resolution slower, no functionality change is expected.  In addition, since plugins are typically cached and read from disk this increase in resolution time should not have significant effect on most workflows.

### DIFF
--- a/changelog/@unreleased/pr-593.v2.yml
+++ b/changelog/@unreleased/pr-593.v2.yml
@@ -1,7 +1,7 @@
 type: improvement
 improvement:
   description: |-
-    If resolvers are specified  for default tasks, they are now used before the default resolver.
+    If resolvers are specified for default tasks, they are now used before the default resolver.
 
     The prior behavior can be maintained by adding the default resolver as the first entry in the resolver list in the config.
   links:
--- a/changelog/@unreleased/pr-593.v2.yml
+++ b/changelog/@unreleased/pr-593.v2.yml
@@ -1,0 +1,8 @@
+type: improvement
+improvement:
+  description: |-
+    If resolvers are specified  for default tasks, they are now used before the default resolver.
+
+    The prior behavior can be maintained by adding the default resolver as the first entry in the resolver list in the config.
+  links:
+    - https://github.com/palantir/godel/pull/593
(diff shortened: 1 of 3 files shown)


changelog/@unreleased/pr-593.v2.yml
```diff
@@ -0,0 +1,18 @@
+type: improvement
+improvement:
+  description: |-
+    Reverse order of computed resolvers list
```
nmiyake, line 4: Was this generated by the app? This content is great for the PR template, but is not the right format/content for the changelog, which is primarily meant to be consumed as release notes by the user.

The release note should focus more on the observable change in behavior rather than on the implementation details -- it should basically say that, if resolvers are specified for defaults tasks, those are now used first before the built-in/default resolver. May also be worth adding that the previous behavior can be maintained after this change by adding the built-in resolver as the first entry in the custom resolver list if specified in config.

nmiyake (changes requested):
Thanks for the PR, the implementation looks great and PR description was also very understandable.

However, request that changelog be updated so that release notes will be easier to consume.

changelog/@unreleased/pr-593.v2.yml
akarnani: Yup, it was autogenerated.  I've updated it to be more succinct.

changelog/@unreleased/pr-593.v2.yml
```diff
@@ -0,0 +1,8 @@
+type: improvement
+improvement:
+  description: |-
+    If resolvers are specified  for default tasks, they are now used before the default resolver.
```
nmiyake, line 4: Typo: the double-space between "specified" and "for" should be a single space.

nmiyake (changes requested):
Thanks for updating! Content looks good, but minor typo.

nmiyake (approved):

nmiyake:
Thanks!

PR #595, opened 2021-06-23
improvement: Upgrade godel-test-plugin to 1.8.0

bmoylan:


nmiyake (approved):
Thanks! Excavator should do this automatically as well, but this does provide better release notes and ensures this happens more quickly.

PR #615, opened 2021-12-07
Use a code-block for the installation command for easy copying

tabboud:
## Before this PR
Code blocks now support an automatic copy utility whereas inline code sections do not. This will prevent users from manually copying and will hopefully reduce a class of developer errors.

## After this PR

==COMMIT_MSG==
Use a code-block for the installation command for easy copying
==COMMIT_MSG==

## Example Gif

https://user-images.githubusercontent.com/5448017/145092684-b8e11c23-3837-4d26-9d4c-531e03aa8bec.mov

### DIFF
--- a/godelinit/README.md
+++ b/godelinit/README.md
@@ -5,7 +5,9 @@ the latest version of gödel to the working directory.
 
 Installation
 ------------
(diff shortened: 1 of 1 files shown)


nmiyake (approved):
Thanks!

PR #696, opened 2023-01-04
Update run-godel-publish GitHub Action to run on release

nmiyake:
## Before this PR
It appears that the behavior of `autorelease-bot` has changed such that, when it creates an autorelease based on a merged PR, it adds the "[skip ci]" directive (see how release 2.72.0 at https://github.com/palantir/godel/commit/03ca0ca8a4bc4298fc4a864dbe9ab19e4ef826e3 has this directive, while release 2.64.0 at https://github.com/palantir/godel/commit/e2ac15753003abf295f79b66d1709984c4752748 does not).

Currently, the GitHub Action logic for publishing this project runs on the "push: tag" event and matches the expected release tag format. However, per https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs, any "push" event workflows are skipped based on this tag.

This makes it such that the GitHub Action no longer runs for this class of autorelease workflows -- it is why the action didn't run for https://github.com/palantir/godel/releases/tag/v2.72.0, and thus didn't produce a release. Timer-based releases still publish properly because they do not include the "skip CI" directive.

## After this PR
Change the GitHub Action logic that runs the publish operation to run on the event that creates releases rather than matching tags on a push event. Ensures that the GitHub Action for releases will be performed even if the release commit has a "[skip ci]" directive. This is also more semantically correct, as it ties the action to the release being created (and should thus theoretically work even if the tag format is changed in the future).

## Possible downsides?
Changes behavior such that if a release is created that doesn't match the previously defined tag format a release will still be performed. Also makes it such that just pushing a version tag will not trigger the publish workflow (a release must specifically be created in GitHub). However, I believe these behavioral changes should be acceptable.

### DIFF
--- a/.github/workflows/publish-godel-artifacts.yml
+++ b/.github/workflows/publish-godel-artifacts.yml
@@ -1,8 +1,8 @@
 name: publish-godel-artifacts
 on:
-  push:
-    tags:
-      - 'v[0-9]+.[0-9]+.[0-9]+**'
+  release:
+    types:
+      - created
 jobs:
   run-godel-publish:
     runs-on: ubuntu-latest


nmiyake:
OK, I've tested this change on a fork at https://github.com/nmiyake/godel and it appears to work. Verified that:

* Verified previous/unmodified behavior: pushing tag that matches regexp with no "[skip ci]" label creates release: https://github.com/nmiyake/godel/releases/tag/v2.72.2 (release is performed by "[github-actions]")
* Workflow does *not* run on pushing tags on commit has "[skip ci]" in commit message
* Syntax is correct for updated workflow
* Creating a release via the UI (https://github.com/nmiyake/godel/releases/tag/v2.72.3, released by "nmiyake") triggers the workflow (even on commit with "[skip ci]" in commit message) and adds download link to the release

Given that the latest release was created by "svc-autorelease" (https://github.com/palantir/godel/releases), I believe that this should work as expected: autorelease-bot should create the release and that should trigger the workflow.

As noted previously, the only behavioral change here is that pushing tags directly will no longer be sufficient to create a release -- in the past, pushing a matching tag would trigger the workflow, which would itself create the release (if it did not already exist) and then upload the artifact. Now, the "manual" workflow would require that, after pushing the tag, a user would need to create a release via the UI or API in order to get the upload to occur. I believe this trade-off is acceptable -- we have primarily moved to using autorelease-bot to always perform releases anyway, and even then we do have a manual fallback (just with an extra step).

tabboud (approved):

PR #706, opened 2023-02-10
Update CircleCI Docker image

nmiyake:
Update CircleCI Docker base image to match based image used by CircleCI Go templates.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ executors:
         type: string
         default: /home/circleci/go/src/github.com/palantir/godel
     docker:
-      - image: cimg/go:1.16-browsers
+      - image: cimg/go:1.18-browsers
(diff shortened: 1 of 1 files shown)


nmiyake:
This is the right thing to do in the abstract as well, but the specific issue that motivated this was that running CGO-compiled binaries in the integration tests was producing the error `/lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found` because the glibc version in the 1.16 image is 2.31.

Given that glibc 2.32 was released 2020-08-05, it seems reasonable to require that. The 1.18 base image uses glibc 2.35.

It would probably be safe to bump the base image to 1.19 or 1.20, but for now I'm keeping at 1.18 since that's the version used in our managed CircleCI templates.

tabboud (approved):

PR #883, opened 2025-05-16
Add cache for plugin and asset configuration

nmiyake:
## After this PR

==COMMIT_MSG==
Adds the ability to cache plugin and asset information when resolving plugins and updates the godel startup logic to use the cache.

The net result is a significant improvement in startup time for projects with multiple plugins and assets.
==COMMIT_MSG==


framework/plugins/plugin.go
```diff
@@ -37,27 +38,62 @@ type pluginInfoWithAssets struct {
...
+		if err := verifyPluginCompatibility(plugins); err != nil {
+			return nil, nil, err
+		}
+
+		// write plugin information to cache file
+		pluginsJSON, err := marshalPluginsInfoJSON(plugins)
+		if err != nil {
+			return nil, nil, errors.Wrapf(err, "failed to marshal plugin information")
+		}
+		if err := writeFileUsingRename(cachePath, pluginsJSON); err != nil {
```
bmoylan, line 94: should we delete outdated cache files too? Otherwise this directory will just grow monotonically

nmiyake: Fair question -- right now, godel does not manage its downloads or cache at all in general. This is true of distributions, plugins, assets, and existing caches for updates. As you point out, this means that the `GODEL_HOME` directory monotonically increases in size over time (and indeed some folks have flagged that they manually prune the directory occasionally to reclaim size).

I think if we did implement cache cleaning/pruning, we would either want to make it an explicit command and/or make it time-based (for the latter, something like "on some defined interval, delete known cache files that have not been modified within some time interval").

My current preference would be notate this/have the cache cleaning as a general work item, but don't view it as critical to be part of this PR, as relatively I think the size shouldn't be too much of a concern (although it's a small sample size, locally I'm seeing size ranges of these cache entries to be between 500 bytes - 10kb -- at the high end of 10kb, that's 100 cache entries to be 1mb).

I will note that the approach taken for this PR is somewhat optimized/predicated on changing the existing implementation as little as possible -- this likely isn't the approach we would take if designing this from scratch (we could probably rearchitect to not have to cache at all or use cache entries that would have more common hits), but the existing logic here is also complicated enough that I don't think it's worth the trade-off.

bmoylan (approved):

PR #886, opened 2025-06-05
Only write plugin cache file if cachePath is non-empty

nmiyake:
## After this PR

==COMMIT_MSG==
Fixes issue where the plugin cache would be attempted to be written even if the cache file path was empty.
==COMMIT_MSG==

### DIFF
--- a/changelog/@unreleased/pr-886.v2.yml
+++ b/changelog/@unreleased/pr-886.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Fixes issue where the plugin cache would be attempted to be written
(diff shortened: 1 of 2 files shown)


nmiyake:
Observed this issue at https://app.circleci.com/pipelines/github/palantir/godel-okgo-asset-golint/873/workflows/95546c03-6a7c-4ab7-82b7-b4a13dd9bfeb/jobs/2986, which invokes code path where `cachePath` is empty (this wasn't caught previously since running `godelw` always sets the cache path to a non-empty value)

bmoylan (approved):

PR #1083, opened 2026-07-10
Update godelw to fix race condition for downloading godel executable

nmiyake:
## After this PR

==COMMIT_MSG==
Ensures that the logic that downloads and moves godel to its distribution directory is resilient to concurrent execution.

Previously, if multiple instances of godelw that resolve to the same version of go were run concurrently in a state where the godel distribution did not exist, the invocation could fail due to a race condition where multiple godelw scripts could download/checksum/remove/move the same directory concurrently.
==COMMIT_MSG==

### DIFF
--- a/resources/wrapper/godelw
+++ b/resources/wrapper/godelw
@@ -155,6 +155,54 @@ function verify_godel_version {
     fi
 }
 
+# Acquires a lock using mkdir (atomic operation).
+# Returns 0 if lock acquired, 1 if timeout.
+function acquire_lock {
+    local lock_dir=$1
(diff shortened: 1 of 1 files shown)


nmiyake:
This should fix the test flakes that we're observing in the repository that tests updating godel versions

tabboud (commented):
Logic mostly looks good. One question though - should we should release the lock after the install is complete rather than hold it for the entire duration of the godel process?

nmiyake:
Thanks, good point -- updated to incorporate suggestion.

tabboud (approved):

